PR #65, opened 2024-11-27
add a feature for dropdown to stretch out to fullscreen on mobile

sun-mota:

## Width and Sizing Behavior

- **Width:** The `auro-dropdown` component will automatically consume the full width of its parent container. To make it narrower, you can style the `bibSizer` part.

- **Styling Options:** Only the following styles can be applied to the `bib` part:
  - `width`
  - `height`
  - `maxWidth`
  - `maxHeight`

## Fullscreen Effect on Mobile View

On mobile view, adding the `mobileFullscreenBreakpoint="{breakpoint-token}"` will switch the dropdown to fullscreen mode. 

## Summary by Sourcery

Add a feature to control dropdown fullscreen behavior on mobile devices using a breakpoint attribute, enhance dropdown styling options, and update documentation accordingly.

New Features:
- Introduce a feature to prevent dropdowns from switching to fullscreen mode on mobile by using the 'mobileFullscreenBreakpoint' attribute.

Enhancements:
- Enhance dropdown styling by allowing specific styles to be applied to the 'bibSizer' part, including width, height, maxWidth, and maxHeight.

Build:
- Add a new build script 'build:dropdown-preset' to extract breakpoint variables for dropdown components.

Documentation:
- Update documentation to include information about the new 'mobileFullscreenBreakpoint' feature and styling options for dropdowns.

## Summary by Sourcery

Add a feature to control dropdown fullscreen behavior on mobile devices using a breakpoint attribute, enhance dropdown styling options, and update documentation accordingly.

New Features:
- Introduce a feature to control dropdown fullscreen behavior on mobile devices using the 'mobileFullscreenBreakpoint' attribute.

Enhancements:
- Enhance dropdown styling by allowing specific styles to be applied to the 'bibSizer' part, including width, height, maxWidth, and maxHeight.

Build:
- Add a new build script 'build:dropdown-preset' to extract breakpoint variables for dropdown components.

Documentation:
- Update documentation to include information about the new 'mobileFullscreenBreakpoint' feature and styling options for dropdowns.


components/dropdown/src/floatingUI.mjs
```diff
@@ -10,7 +10,35 @@ export default class AuroFloatingUI {
     this.keyDownHandler = null;
   }
 
+  mirrorSize(fullscreen) {
```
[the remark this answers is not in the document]
sun-mota: fullscreen styling in javascript is necessary because there is a logic that manipulate bibContent's inline style on non-mobile screen, so style from css will get ignored.

sun-mota:
@sourcery-ai review

components/dropdown/src/floatingUI.mjs
```diff
@@ -10,7 +10,39 @@ export default class AuroFloatingUI {
...
+      }
+      bibContent.style.maxWidth = sizerStyle.maxWidth;
+      bibContent.style.maxHeight = sizerStyle.maxHeight;
+    }
+  }
+
   position() {
+    // fullscreen on mobileView 
+    if (this.element.bib.mobileFullscreenBreakpoint) {
+      const isMobile = window.matchMedia(`(max-width: ${this.element.bib.mobileFullscreenBreakpoint})`).matches;
```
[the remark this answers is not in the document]
sun-mota: this requires a big renovation, can be addressed in a separate ticket later.

components/dropdown/src/auro-dropdown.js
```diff
@@ -37,6 +40,7 @@ import './auro-dropdownBib.js';
  * @attr { Boolean } noToggle - If declared, the trigger will only show the the dropdown bib.
  * @attr { Boolean } focusShow - if declared, the the bib will display when focus is applied to the trigger.
  * @attr { Boolean } noHideOnThisFocusLoss - If declared, the dropdown will not hide when moving focus outside the element.
+ * @attr { String } mobileFullscreenBreakpoint - Defines the screen size breakpoint (`lg`, `md`, `sm`, or `xs`) at which the dropdown switches to fullscreen mode on mobile. The dropdown will automatically display in fullscreen mode when the screen size is equal to or smaller than the selected breakpoint.
```
jason-capsule42, line 43:
```suggestion
 * @attr { String } mobileFullscreenBreakpoint - Defines the screen size breakpoint (`lg`, `md`, `sm`, or `xs`) at which the dropdown switches to fullscreen mode on mobile. When expanded, the dropdown will display in fullscreen mode if the screen size is equal to or smaller than the selected breakpoint.
```

components/dropdown/src/auro-dropdown.js
```diff
@@ -229,14 +237,19 @@ export class AuroDropdown extends LitElement {
 
   firstUpdated() {
     this.floater.configure(this);
+    this.bibContent = this.floater.element.bib;
+    if (this.mobileFullscreenBreakpoint) {
+      const docStyle = getComputedStyle(document.documentElement);
+      const breakpointValue = docStyle.getPropertyValue(DESIGN_TOKEN_BREAKPOINT_PREFIX + this.mobileFullscreenBreakpoint);
+      // if (!breakpointValue) {
+      //   console.warn('There is no breakpoint token called', DESIGN_TOKEN_BREAKPOINT_PREFIX + this.mobileFullscreenBreakpoint);
+      // }
```
jason-capsule42, line 246: Remove commented out code

components/dropdown/src/auro-dropdown.js
```diff
@@ -229,14 +237,19 @@ export class AuroDropdown extends LitElement {
 
   firstUpdated() {
     this.floater.configure(this);
+    this.bibContent = this.floater.element.bib;
+    if (this.mobileFullscreenBreakpoint) {
+      const docStyle = getComputedStyle(document.documentElement);
+      const breakpointValue = docStyle.getPropertyValue(DESIGN_TOKEN_BREAKPOINT_PREFIX + this.mobileFullscreenBreakpoint);
```
jason-capsule42, line 243: I think we need to make sure that `this.mobileFullScreenBreakpoint` is a valid value before we run this code. Otherwise, we will try to do getPropertyValue on something invalid and pass that down to the the bib.

jason-capsule42: Additional:

I think auro-dropdown should just pass the breakpoint string (sm, md, etc..) down to auro-dropdownbib and all the functional code can live there.

Additionally, rather than doing this on firstUpdated this should run inside:

```
updated(changedProperties) {
  if (changedProperties.has('mobileFullscreenBreakpoint') {
    // execute the code here...
  }
}
```

That way we handle the scenario where a consumer might hypothetically change the breakpoint to trigger mobile after the component get's loaded into the DOM.

components/dropdown/src/auro-dropdown.js
```diff
@@ -361,6 +374,7 @@ export class AuroDropdown extends LitElement {
         <div class="slotContent">
           <slot @slotchange="${this.handleDefaultSlot}"></slot>
         </div>
+        <div id="bibSizer" part="bib" style="position: absolute; pointer-events: none; opacity: 0; z-index: -1;"></div>
```
jason-capsule42, line 377: All the inline styles should get moved into the style.scss file under a class.

jason-capsule42 (changes requested):

jason-capsule42 (changes requested):

components/dropdown/src/styles/bibStyles.scss
```diff
@@ -11,15 +11,32 @@
...
+
+ .container {
+    top: 0;
+    left: 0;
+    width: 100dvw;
+    max-width: none;
+    height: 100dvh;
+    max-height: none;
+ }
 }
```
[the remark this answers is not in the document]
sun-mota: For the second part in floatingUI.mjs, 
`position` function returns before calling `computePosition`, since  there is no need to event compute the position in fullscreen mode.

components/dropdown/src/floatingUI.mjs
```diff
@@ -10,22 +10,53 @@ export default class AuroFloatingUI {
...
+    }
+
+    this.handleMobileFullscreen(false);
+    this.mirrorSize(false);
+
     // Define the middlware for the floater configuration
     const middleware = [
       offset(this.element.floaterConfig.offset || 0),
+      ...(this.element.floaterConfig.flip ? [flip()] : []), // Add flip middleware if flip is enabled
+      ...(this.element.floaterConfig.autoPlacement ? [autoPlacement()] : []), // Add autoPlacement middleware if autoPlacement is enabled
```
jason-capsule42, line 57: I would have thought that the falsy statement of the ternary functions would have reset the array to be empty rather than just not adding the configuration. Did you validate that this for sure works correctly?

components/dropdown/src/styles/style.scss
```diff
@@ -16,6 +16,13 @@
   position: relative;
 }
 
+#bibSizer {
+  position: absolute;
+  z-index: -1;
```
jason-capsule42, line 21: Do we need this? It's not an element that you can focus, right?

jason-capsule42: In fact... do we need  the pointer events rule either?

components/dropdown/src/styles/bibStyles.scss
```diff
@@ -11,23 +11,39 @@
 
 /* stylelint-disable no-descending-specificity, selector-max-attribute */
 
+:host {
+  position: absolute;
+  z-index: 1;
```
jason-capsule42, line 16: Maybe we can move these two rules into this select `:host([data-show])` since they don't do anything until then. What do you think?

components/dropdown/src/floatingUI.mjs
```diff
@@ -265,8 +307,11 @@ export default class AuroFloatingUI {
   configure(elem) {
     this.element = elem;
     this.element.trigger = this.element.shadowRoot.querySelector('#trigger');
-    this.element.bib = this.element.shadowRoot.querySelector('#bib');
     this.element.triggerChevron = this.element.shadowRoot.querySelector('#showStateIcon');
+    this.element.bib = this.element.shadowRoot.querySelector('#bib');
+    this.element.bibSizer = this.element.shadowRoot.querySelector('#bibSizer');
+    
+    document.body.append(this.element.bib);
```
jason-capsule42, line 314: I am confused why this was added. Help me understand?

components/dropdown/src/floatingUI.mjs
```diff
@@ -38,6 +69,17 @@ export default class AuroFloatingUI {
     });
   }
 
+  handleMobileFullscreen(isMobile) {
```
jason-capsule42, line 72: This function just positions the modal without affecting size. Meaning, it doesn't do everything you might think of when you say "handleMobileFullscreen". Let's give this a more semantically correct function name. Something along the lines of `positionFullScreenBib()` or something to that affect.

components/dropdown/src/floatingUI.mjs
```diff
@@ -10,22 +10,53 @@ export default class AuroFloatingUI {
...
+
+  getPositioningStrategy() {
+    if (this.element.bib.mobileFullscreenBreakpoint) {
+      const isMobile = window.matchMedia(`(max-width: ${this.element.bib.mobileFullscreenBreakpoint})`).matches;
+      return isMobile ? 'fullscreen' : 'floating';
+    }
+    return 'floating';
+  }
+
   position() {
```
jason-capsule42, line 42: JSDoc documentation needed for this function and looks like it should be marked private.

components/dropdown/src/floatingUI.mjs
```diff
@@ -10,22 +10,53 @@ export default class AuroFloatingUI {
...
+    const strategy = this.getPositioningStrategy();
+    if (strategy === 'fullscreen') {
+      this.handleMobileFullscreen(true);
+      this.mirrorSize(true);
+      return;
```
jason-capsule42, line 47: We aren't actually returning anything in this function and the JSDoc syntax likely won't be happy once you add but have nothing to actually return in the code. I think it would be better if we use an `else {...}` here rather than a `return` to force an execution break.

components/dropdown/src/floatingUI.mjs
```diff
@@ -10,22 +10,53 @@ export default class AuroFloatingUI {
...
+      }
+      if (sizerStyle.height !== '0px') {
+        bibContent.style.height = sizerStyle.height;
+      }
+      bibContent.style.maxWidth = sizerStyle.maxWidth;
+      bibContent.style.maxHeight = sizerStyle.maxHeight;
+    }
+  }
+
+  getPositioningStrategy() {
```
jason-capsule42, line 34: Need to add JSDoc documentation for the function and mark it private.

components/dropdown/src/floatingUI.mjs
```diff
@@ -10,22 +10,53 @@ export default class AuroFloatingUI {
...
+    if (this.element.bib.mobileFullscreenBreakpoint) {
+      const isMobile = window.matchMedia(`(max-width: ${this.element.bib.mobileFullscreenBreakpoint})`).matches;
+      return isMobile ? 'fullscreen' : 'floating';
+    }
+    return 'floating';
+  }
```
jason-capsule42, line 40:
```suggestion
    let strategy = 'floating';
    if (this.element.bib.mobileFullscreenBreakpoint) {
      const isMobile = window.matchMedia(`(max-width: ${this.element.bib.mobileFullscreenBreakpoint})`).matches;
      if (isMobile) {
        strategy = 'fullscreen';
      }
    }
    return strategy;
  }
```

this way you only define the string `floating` once and future maintenance is a little easier.

components/dropdown/src/floatingUI.mjs
```diff
@@ -10,22 +10,53 @@ export default class AuroFloatingUI {
     this.keyDownHandler = null;
   }
 
+  mirrorSize(fullscreen) {
```
jason-capsule42, line 13: Add JSDoc documentation and set to private

components/dropdown/src/auro-dropdownBib.js
```diff
@@ -40,6 +50,21 @@ export class AuroDropdownBib extends LitElement {
...
+          // verify the defined breakpoint is valid and exit out if not
+          const validatedValue = DESIGN_TOKEN_BREAKPOINT_OPTIONS.includes(value) ? value : undefined;
+          if (!validatedValue) {
+            return undefined;
+          }
+
+          // get the pixel value for the defined breakpoint
+          const docStyle = getComputedStyle(document.documentElement);
+          return docStyle.getPropertyValue(DESIGN_TOKEN_BREAKPOINT_PREFIX + value);
+        },
```
jason-capsule42, line 66: I wonder if this should be moved to a function to keep the constructor clean. Thoughts?

```suggestion
        converter: getBreakpointValue()
```

components/dropdown/src/auro-dropdownBib.js
```diff
@@ -10,10 +10,20 @@ import styleCss from "./styles/bibStyles-css.js";
 import colorCss from "./styles/bibColors-css.js";
 import tokensCss from "./styles/tokens-css.js";
 
+
+const DESIGN_TOKEN_BREAKPOINT_PREFIX = '--ds-grid-breakpoint-';
```
jason-capsule42, line 14: I wonder if there is a way we can future proof this to reduce the risk we break the code if the breakpoint token names change in the future. Let's open an investigation token to consider this as a future possible improvement.

components/dropdown/src/auro-dropdown.js
```diff
@@ -361,12 +363,14 @@ export class AuroDropdown extends LitElement {
...
         </div>
+        <div id="bibSizer" part="size"></div>
         <auro-dropdownbib
           id="bib"
           role="tooltip"
           ?common="${this.common}"
           ?rounded="${this.common || this.rounded}"
-          ?inset="${this.common || this.inset}">
+          ?inset="${this.common || this.inset}"
+          mobileFullscreenBreakpoint="${this.mobileFullscreenBreakpoint}">
```
jason-capsule42, line 373:
```suggestion
          .mobileFullscreenBreakpoint="${this.mobileFullscreenBreakpoint}">
```

Adding the `.` should make it so the attribute is only added if there is a value for it.

jason-capsule42 (changes requested):

components/dropdown/src/auro-dropdown.js
jason-capsule42: To clarify,... using `.` is for string attributes and `?` is for boolean attributes. They affectively do the same thing in that they only add the attribute when there is a not a false/null/undefined value to apply.

components/dropdown/src/floatingUI.mjs
sun-mota: yes, it's validated. The ternary operator doesn’t reset or clear arrays automatically. because each line is to lay out the values of the array with `...()`, not directly setting any value.

components/dropdown/src/styles/style.scss
sun-mota: `z-index:-1` will help the performance by putting the layer to be the most back so that other layer would get any cursor event. `pointer-events` rule would also help the performance with bypassing the cursor event on this layer. 
IMO, I think adding these rule wont hurt and rather help a little bit.

components/dropdown/src/styles/bibStyles.scss
sun-mota: Changing `position` style would make the browser to re-calculate of the layer's position, which affects the performance. I'd not recommend.

components/dropdown/src/floatingUI.mjs
sun-mota: this is the part of the floatingUI's logic which takes #bib out of the dropdown and teleport to the body

components/dropdown/src/auro-dropdownBib.js
sun-mota: created #81

components/dropdown/src/floatingUI.mjs
sun-mota: this line is moved from `auro-dropdown.js` to consolidate all floating logic to be in `floatingUI.mjs`

jason-capsule42 (approved):

PR #116, opened 2024-12-16
fix(auro-menu): ensure correct aria-selected states for unselected menuoptions

chrisfalaska:

- Ensure correct `aria-selected` states for unselected `auro-menu` options by setting `aria-selected` to `false` when a menu `option` is not selected.

Issues originally opened in: https://github.com/AlaskaAirlines/auro-menu/issues/187

## Updates

- Initialize `aria-selected` attribute to `false` during component connection
- Added an `update` observer for `aria-selected` attributes when selected property changes
- Changed removal of `aria-selected`, setting it to `false` when clearing selections

## Testing
Screenshot 2024-12-16 at 12 09 14 PM

https://github.com/user-attachments/assets/7077290d-0280-4fc2-a31d-a9408725dc45

### DIFF
--- a/components/menu/src/auro-menu.js
+++ b/components/menu/src/auro-menu.js
@@ -185,7 +185,6 @@ export class AuroMenu extends LitElement {
       this.items.forEach((item) => {
         item.classList.remove('active');
         item.removeAttribute('selected');
-        item.setAttribute('aria-selected', 'false');
       });
     }
   }
--- a/components/menu/src/auro-menu.js
+++ b/components/menu/src/auro-menu.js
@@ -185,7 +185,7 @@ export class AuroMenu extends LitElement {
       this.items.forEach((item) => {
         item.classList.remove('active');
         item.removeAttribute('selected');
-        item.removeAttribute('aria-selected');
+        item.setAttribute('aria-selected', 'false');
       });
     }
   }
--- a/components/menu/demo/api.min.js
+++ b/components/menu/demo/api.min.js
@@ -666,7 +666,7 @@ class AuroMenuOption extends r {
     });
   }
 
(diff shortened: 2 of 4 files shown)


components/menu/src/auro-menu.js
```diff
@@ -185,7 +185,7 @@ export class AuroMenu extends LitElement {
       this.items.forEach((item) => {
```
sun-mota, line 185: looks like `this.selected = false;` would be needed.

components/menu/src/auro-menu.js
```diff
@@ -185,7 +185,7 @@ export class AuroMenu extends LitElement {
...
         item.classList.remove('active');
         item.removeAttribute('selected');
-        item.removeAttribute('aria-selected');
+        item.setAttribute('aria-selected', 'false');
```
sun-mota, line 188: with the line above, line 188 should not be needed as menuoption's updated function should take care of this.

chrisfalaska: Removed

components/menu/src/auro-menu.js
chrisfalaska: `this` would be the `auro-menu` component, but not the `menuoption`.

Should it be `item.selected = false;`?

```
resetOptionsStates() {
  this.optionSelected = undefined;
  if (this.items) {
    this.items.forEach((item) => {
      item.classList.remove('active');
      item.removeAttribute('selected');
      item.selected = false;
    });
  }
}
````

sun-mota: Oh my comment was mixed up. I think `item.removeAttribute('selected');` would do it enough.

sun-mota (approved):

PR #133, opened 2024-12-19
feat(menu): add multi-select

chrisfalaska:


components/combobox/demo/api.min.js
```diff
@@ -8065,7 +8065,8 @@ var tokensCss$1 = i$c`:host{--ds-auro-menu-divider-color: var(--ds-color-border-
  * @attr {Boolean} disabled - When true, the entire menu and all options are disabled;
  * @attr {Boolean} noCheckmark - When true, selected option will not show the checkmark.
  * @attr {Boolean} loading - When true, displays a loading state using the loadingIcon and loadingText slots if provided.
- * @attr {String} value - Value selected for the menu.
+ * @attr {Boolean} multiSelect - When true, the selected option can be multiple option.
+ * @attr {String} value - Value selected for the menu, can be String or Array if `multiSelect` is true.
```
jason-capsule42, line 8069:
```suggestion
 * @attr {String || Array} value - Value selected for the menu, can be String or Array if `multiSelect` is true.
```

Does this work?

components/combobox/demo/api.min.js
```diff
@@ -8093,6 +8094,7 @@ class AuroMenu extends r$6 {
     this.noCheckmark = false;
     this.optionActive = undefined;
     this.loading = false;
+    this.multiSelect = undefined;
```
jason-capsule42, line 8097: Shouldn't this start as false like noCheckmark and loading?

components/menu/src/auro-menu.js
```diff
@@ -212,10 +215,55 @@ export class AuroMenu extends LitElement {
...
+   * Toggle the selection state of the menuoption.
+   * @param {Object} option - The menuoption to toggle the selection state.
+   * @private
+   * @return {void}
+   */
+  toggleSelectionState(option) {
+    if (option.hasAttribute('selected')) {
+      this.handleDeselectState(option);
+    } else {
+      this.handleLocalSelectState(option);
```
jason-capsule42, line 233: I am a little confused on the semantic naming of these functions. Why is one "Local" and the other isn't? Does that imply something?

jason-capsule42 (changes requested):
Refactor selection logic based on our conversation so that the value/optionSelected logic is isolated from the UI updates. And make the UI updates responsive to value/optionSelected in the updated() function rather than triggered directly by the selection event.

components/combobox/demo/api.min.js
chrisfalaska: Fixed

components/menu/src/auro-menu.js
chrisfalaska: It was driving me crazy as well, I have changed them.

components/combobox/demo/api.min.js
chrisfalaska: Strangely, this is Combobox. Not sure how it found its way into this PR. Perhaps an artifact of a rebase?

chrisfalaska:
sourcery check

chrisfalaska:
@sourcery-ai review

components/menu/docs/partials/api.md
```diff
@@ -74,6 +74,23 @@ The `auro-menu` component supports the use of the `matchWord` attribute to highl
 
 </auro-accordion>
 
+### Multi Select<a name="multiSelect"></a>
+
+The `auro-menu` supports a multi-select option. To use, place the `multiSelect` attribute on the `<auro-menu>` element tag. When applied, the `value` attribute will become an Array versus String value.
```
jason-capsule42, line 79: Attribute names are case insensitive so we should always document their use as fully lowercase.

components/menu/docs/api.md
```diff
@@ -30,17 +29,17 @@ The auro-menu element provides users a way to select from a list of options.
...
 | `auroMenu-selectedOption`     | `CustomEvent<any>`                               | Notifies that a new menuoption selection has been made. |
 
 ## Slots
 
 | Name          | Description                                 |
 |---------------|---------------------------------------------|
-|               | Slot for insertion of menu options.         |
 | `loadingIcon` | Icon to show while loading attribute is set |
 | `loadingText` | Text to show while loading attribute is set |
+| `slot`        | Slot for insertion of menu options.         |
```
jason-capsule42, line 42: Why was this changed? You don't need a name for the default slot, was this auto-generated?

components/menu/docs/api.md
```diff
@@ -6,22 +6,21 @@ The auro-menu element provides users a way to select from a list of options.
...
+| `optionSelected`        | `optionSelected` | `object`  | "undefined" | Specifies the current selected menuOption.       |
+| `value`                 | `value`          | `string`  | "undefined" | Value selected for the menu. Can be undefined, a string, or an array of strings if `multiSelect` is true. |
 
 ## Methods
 
-| Method               | Type                            | Description                                      |
-|----------------------|---------------------------------|--------------------------------------------------|
-| `makeSelection`      | `(): void`                      | Process actions for making making a menuoption selection. |
-| `resetOptionsStates` | `(): void`                      | Reset the menu and all options.                  |
-| `selectNextItem`     | `(moveDirection: string): void` | Using value of current this.index evaluates index<br />of next :focus to set based on array of this.items ignoring items<br />with disabled attr.<br /><br />The event.target is not used as the function needs to know where to go,<br />versus knowing where it is.<br /><br />**moveDirection**: Up or Down based on keyboard event. |
```
jason-capsule42, line 24: I'm unclear why these functions were removed.

components/menu/docs/api.md
chrisfalaska: No, I thought it was a typo because it appeared blank in the docs. I'll revert it.

components/menu/docs/api.md
chrisfalaska: They're there but have been refactored and renamed:

`makeSelection` - This is still present but now handles single and multi-select.

`resetOptions` - This has been split into two separate functions: `clearSelection` which handles just the state and `reset` which is more robust and handles state, events & ui.

`selectNextItem` - this has been renamed to ` navigateOptions(direction)`.

components/menu/docs/partials/api.md
chrisfalaska: Great callout.

components/menu/docs/api.md
chrisfalaska: I noticed with the exception of `reset`, they are listed as `@private`. Should they be exposed?

jason-capsule42: Only if the other components (select/combobox) need to call them.

components/menu/src/auro-menu.js
```diff
@@ -106,198 +216,344 @@ export class AuroMenu extends LitElement {
...
+  /**
+   * Updates menu state when an option is selected.
+   * @private
+   * @param {HTMLElement} option - The option element to select.
+   */
+  handleSelectState(option) {
+    if (this.multiSelect) {
+      // Initialize arrays if undefined (first interaction)
+      const currentValue = Array.isArray(this.value) ? this.value : [];
+      const currentSelected = Array.isArray(this.optionSelected) ? this.optionSelected : [];
```
sun-mota, line 329: I think these line are doing the same as `AuroMenu.processToArray`. then it's good have a good use of that function.

components/menu/src/auro-menu.js
```diff
@@ -106,198 +216,344 @@ export class AuroMenu extends LitElement {
...
 
     this.loadingSlots = this.querySelectorAll("[slot='loadingText'], [slot='loadingIcon']");
+    this.initializeMenu();
   }
 
   updated(changedProperties) {
-    if (changedProperties.has('matchWord')) {
-      this.markOptions();
+    // Handle preset values
+    if ((changedProperties.has('value') || changedProperties.has('optionSelected')) &&
```
sun-mota, line 253: we might need a validator to check if the value is set as `non-array` when `multiselect` is true. Then, food for though, with the value type validation, should menu auto-correct it to be array or throw an error?

components/menu/src/auro-menu.js
```diff
@@ -94,6 +149,61 @@ export class AuroMenu extends LitElement {
...
+  }
+
+  // Compares array for changes
+  static arrayOrUndefinedHasChanged(newVal, oldVal) {
+    const newArray = AuroMenu.processToArray(newVal);
+    const oldArray = AuroMenu.processToArray(oldVal);
+
+    return !AuroMenu.arraysAreEqual(newArray, oldArray);
+  }
+
```
sun-mota, line 206: These are very good utility functions but would be better to be separated out from Menu.

components/menu/src/auro-menu.js
```diff
@@ -94,6 +149,61 @@ export class AuroMenu extends LitElement {
...
+      return arr1 === arr2;
+    }
+
+    // If arrays have different lengths, they are not equal
+    if (arr1.length !== arr2.length) {
+      return false;
+    }
+
+    // If every item at each index is the same, return true
+    return arr1.every((item, index) => item === arr2[index]);
```
sun-mota, line 196: would `JSON.stringfy` be better performing?

components/menu/src/auro-menu.js
```diff
@@ -106,198 +216,344 @@ export class AuroMenu extends LitElement {
...
-          this.hasLoadingPlaceholder
-        }
-      });
-      this.setAttribute("aria-busy", this.hasAttribute("loading"));
-      this.dispatchEvent(event);
+      this.updateLoadingState();
+    }
+
+    if (changedProperties.has('noCheckmark')) {
+      this.propagateNoCheckmark();
```
sun-mota, line 290: Each of these functions are running loops for each items. 
If there are `n` properties changed with `m` items, then it will be O(n*m) complexity. Would there be a way that we can consolidate updating items with the minimal loop?

sun-mota (changes requested):

components/menu/src/auro-menu.js
chrisfalaska: Agreed! I'm not sure if we have such a utility concept for Auro? Library, maybe?

components/menu/src/auro-menu.js
chrisfalaska: I just tried both and updated with a for loop

components/menu/src/auro-menu.js
sun-mota: yes, probably Library.

components/menu/src/auro-menu.js
chrisfalaska: Great point.

I consolidated them into a single loop in `updateItemsState`

components/menu/src/auro-menu.js
chrisfalaska: Fixed!

components/menu/src/auro-menu.js
chrisfalaska: I moved them into their own file called `/src/auro-menu-utils.js`.

Perhaps we can keep them there, or potentially move them to Library in another PR.

components/menu/src/auro-menu.js
chrisfalaska: Great idea. I added a `validateArrayValue` util. I think throwing an error would be best to allow the consumer to understand the API and make the correct changes.

chrisfalaska:
@sourcery-ai review

components/menu/src/auro-menu.js
```diff
@@ -106,198 +175,308 @@ export class AuroMenu extends LitElement {
...
+    this.removeEventListener('mousedown', this.handleMouseSelect);
+    this.removeEventListener('auroMenuOption-mouseover', this.handleOptionHover);
+    this.removeEventListener('slotchange', this.handleSlotChange);
+
+    super.disconnectedCallback();
   }
 
   firstUpdated() {
-    // Add the tag name as an attribute if it is different than the component name
     this.runtimeUtils.handleComponentTagRename(this, 'auro-menu');
```
sun-mota, line 199: nit, this can be `AuroLibraryRuntimeUtils.prototype.handleComponentTagRename` then `this.runtimeUtils` wont be needed.

components/menu/src/auro-menu-utils.js
```diff
@@ -0,0 +1,112 @@
...
+ * @param {Array|any} oldVal - Old value to compare
+ * @returns {boolean} True if arrays have changed
+ */
+export function arrayOrUndefinedHasChanged(newVal, oldVal) {
+  const newArray = processToArray(newVal);
+  const oldArray = processToArray(oldVal);
+
+  return !arraysAreEqual(newArray, oldArray);
+}
+
```
sun-mota, line 81: I'd recommend having a separate file only for Array related functions as it'll be very useful for other components in the future.
`isOptionInteractive` and `dispatchMenuEvent` seems less universal.

jordanjones243 (dismissed):
Can there be some cleanup around commits? 72 commits for 10 file changes seems a bit excessive and I see some duplicate commits in there.

components/menu/src/auro-menu-utils.js
chrisfalaska: I agree, I think the array functions could be helpful in Library. But for the this PR, what do you think about maintaining all of the relevant auro-menu utils in a separate file?

In general, I'd prefer if we went more in this direction with Auro components, separating concerns into partials. The main component file is 700 lines of code, which I find hard to work with.

sun-mota: Considering moving array functions to Library later, I'd have only array functions in a separate file like `array-util.js` and leave other menu specific util functions in `auro-menu-util.js`
I agree 700 lines of code is too massive.

chrisfalaska:
> Can there be some cleanup around commits? 72 commits for 10 file changes seems a bit excessive and I see some duplicate commits in there.

I can squash and merge into a single commit. It's one big refactor, I'd argue preserving the commit history is not so important.

components/menu/src/auro-menu-utils.js
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  // Return the value if it is already an array
+  if (Array.isArray(value)) {
+    return value;
+  }
+
+  try {
+    // If value is a JSON string, parse it
+    const parsed = typeof value === 'string' ? JSON.parse(value) : value;
```
sun-mota, line 30: nit; we know value is not Array already from line 24. maybe we can throw an error when the type is not string at this level directly?

components/menu/src/auro-menu.js
```diff
@@ -106,198 +169,308 @@ export class AuroMenu extends LitElement {
...
+      // Get the initial value from the attribute
+      const initialValue = this.getAttribute('value');
+
+      // Find the corresponding option in the menu items
+      const matchingOption = this.items.find((item) => item.value === initialValue);
+
+      if (matchingOption) {
+        // Convert to array and select the matching option (single-select mode)
+        this.value = processToArray(initialValue);
+        this.optionSelected = processToArray(matchingOption);
```
sun-mota, line 218: processToArray is not needed because these value are already validated to be array or undefined with converter.

components/menu/src/auro-menu.js
```diff
@@ -106,198 +169,308 @@ export class AuroMenu extends LitElement {
...
-    if (this.items) {
-      this.items.forEach((item) => {
-        item.classList.remove('active');
-        item.removeAttribute('selected');
-      });
+  handleSelectState(option) {
+    if (this.multiSelect) {
+      // Initialize arrays if undefined (first interaction)
+      const currentValue = processToArray(this.value) || [];
+      const currentSelected = processToArray(this.optionSelected) || [];
```
sun-mota, line 339: processToArray is not needed because these value are already validated to be array or undefined with converter.

sun-mota (changes requested):
select does not show selected item properly.
combobox does not work.

components/menu/src/auro-menu.js
```diff
@@ -106,198 +169,308 @@ export class AuroMenu extends LitElement {
...
   firstUpdated() {
-    // Add the tag name as an attribute if it is different than the component name
-    this.runtimeUtils.handleComponentTagRename(this, 'auro-menu');
+    AuroLibraryRuntimeUtils.prototype.handleComponentTagRename(this, 'auro-menu');
 
-    this.addEventListener('keydown', this.handleKeyDown);
+    // Make the menu focusable
+    if (!this.hasAttribute('tabindex')) {
+      this.setAttribute('tabindex', '0');
+    }
```
jason-capsule42, line 198: This needs to be removed.

components/menu/src/auro-menu.js
```diff
@@ -307,222 +480,198 @@ export class AuroMenu extends LitElement {
...
-   * @private
-   */
-  getSelectedIndex() {
-    // find the first `selected` and not `disabled`, `hidden` or `static` option
-    const index = this.items.findIndex((option) => option.hasAttribute('selected') && this.optionInteractive(option));
+    const option = this.items[this.index];
+    if (option && !option.hasAttribute('disabled')) {
+      if (this.isOptionSelected(option)) {
+        // If the option is already selected, deselect it
+        this.handleDeselectState(option);
```
jason-capsule42, line 495: This deselect should only happen when the menu is `multiselect` when more than one option is already selected.

e.g. clicking options is not a valid way to `reset` the menu to an initial state.

components/combobox/src/auro-combobox.js
```diff
@@ -345,13 +383,19 @@ export class AuroCombobox extends LitElement {
 
     // handle the menu event for an option selection
     this.menu.addEventListener('auroMenu-selectedOption', () => {
+      // dropdown bib should hide when making a selection
+      this.hideBib();
+
       if (this.menu.optionSelected) {
-        if (this.optionSelected !== this.menu.optionSelected) {
-          this.optionSelected = this.menu.optionSelected;
+        const selected = Array.isArray(this.menu.optionSelected) ? this.menu.optionSelected[0] : this.menu.optionSelected;
```
sun-mota, line 390: we already know menu.optionSelected is an array. any reason to check again?

components/combobox/src/auro-combobox.js
```diff
@@ -415,8 +456,60 @@ export class AuroCombobox extends LitElement {
         this.validation.validate(this);
       }
 
-      if (typeof this.value === 'object') {
-        this.value = '';
+      // Set to undefined if empty
+      if (Array.isArray(this.value) && this.value.length === 0) {
```
sun-mota, line 460: same here. we already know the value should be an array

components/combobox/src/auro-combobox.js
```diff
@@ -415,8 +456,60 @@ export class AuroCombobox extends LitElement {
...
+        this.value = undefined;
+      }
+    });
+
+    this.input.addEventListener('input', () => {
+      this.menu.matchWord = this.input.value;
```
sun-mota, line 466: this part was removed from one of the merged PRs.

components/combobox/src/auro-combobox.js
```diff
@@ -454,14 +547,11 @@ export class AuroCombobox extends LitElement {
    * @returns {void}
    */
   handleInputValueChange() {
-    this.menu.matchWord = this.input.value;
-    this.optionActive = null;
+    // Store value as array or undefined
```
sun-mota, line 550: this function is rewritten in a merged PRs and all the logics are removed.

components/select/src/auro-select.js
```diff
@@ -460,15 +505,21 @@ export class AuroSelect extends LitElement {
...
   }
 
-  updated(changedProperties) {
-    // After the component is ready, send direct value changes to auro-menu.
+  async updated(changedProperties) {
     if (changedProperties.has('value')) {
       if (this.value) {
-        this.menu.value = this.value;
+        // Ensure menu always receives array of values
+        this.menu.value = Array.isArray(this.value) ? this.value : [this.value];
```
sun-mota, line 517: any reason to check the Array type?

sun-mota (changes requested):

components/menu/src/auro-menu.js
```diff
@@ -106,198 +167,329 @@ export class AuroMenu extends LitElement {
...
-      this.items.forEach((item) => {
-        item.classList.remove('active');
-        item.removeAttribute('selected');
-      });
+  handleSelectState(option) {
+    if (this.multiSelect) {
+      const currentValue = this.value || [];
+      const currentSelected = this.optionSelected || [];
+
+      if (!currentValue.includes(option.value)) {
```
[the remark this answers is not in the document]
chrisfalaska: If we remove the check, do you think it will be clear enough that the function can only be used if `optionSelected` is checked first? Does it make it less useful? 

That would mean `handleSelectState` could only be safely called after another check, etc.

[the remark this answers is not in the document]
chrisfalaska: Kept check in after speaking with @sun-mota .

Simplified makeSelection with toggleOption.

sun-mota (approved):

PR #134, opened 2024-12-19
dropdown/select: when there is no placeholder, label will have default font-size

sun-mota:

## Summary by Sourcery

Update the dropdown and select components to use the default font size for labels when no trigger content is present. Update the default placeholder for auro-select to an empty string.

### DIFF
--- a/components/combobox/demo/index.min.js
+++ b/components/combobox/demo/index.min.js
@@ -2791,7 +2791,7 @@ let AuroIcon$2 = class AuroIcon extends BaseIcon$2 {
 
 var iconVersion$2 = '6.1.1';
 
-var styleCss$1$2 = i$3$1`:host{position:relative;display:inline-block;max-width:100%}:host([fluid]){display:block}#bibSizer{position:absolute;z-index:-1;opacity:0;pointer-events:none}.label{font-size:var(--ds-text-body-size-xs, 0.75rem);line-height:var(--ds-text-body-size-default, 1rem);white-space:normal}.trigger{position:relative;display:flex;align-items:center;border-width:1px;border-style:solid}@media(hover: hover){.trigger:hover{cursor:pointer}}.triggerContentWrapper{overflow:hidden;flex:1;text-overflow:ellipsis;white-space:nowrap}#showStateIcon{display:flex;height:100%;align-items:center;margin-left:var(--ds-size-100, 0.5rem)}#showStateIcon [auro-icon]{height:var(--ds-size-300, 1.5rem);line-height:var(--ds-size-300, 1.5rem)}#showStateIcon[data-expanded=true] [auro-icon]{transform:rotate(-180deg)}.helpText{margin-top:var(--ds-size-50, 0.25rem);font-size:var(--ds-text-body-size-xs, 0.75rem);line-height:var(--ds-text-body-size-default, 1rem)}:host([matchwidth]) #bibSizer{width:100%}:host([disabled]){pointer-events:none}:host([inset]) .trigger{padding:var(--ds-size-150, 0.75rem) var(--ds-size-200, 1rem)}:host([common]) .trigger,:host([inset][bordered]) .trigger{padding:var(--ds-size-200, 1rem) var(--ds-size-150, 0.75rem)}:host([common]) .trigger,:host([rounded]) .trigger{border-radius:var(--ds-border-radius, 0.375rem)}`;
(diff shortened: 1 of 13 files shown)


components/dropdown/src/auro-dropdown.js
```diff
@@ -236,6 +277,12 @@ export class AuroDropdown extends LitElement {
     if (changedProperties.has('mobileFullscreenBreakpoint')) {
       this.bibContent.mobileFullscreenBreakpoint = this.mobileFullscreenBreakpoint;
     }
+
+    if (this.triggerContentSlot) {
```
jordanjones243, line 281: Do we need to do this check EVERY time the component gets updated?

components/dropdown/src/auro-dropdown.js
```diff
@@ -236,6 +277,12 @@ export class AuroDropdown extends LitElement {
...
+      this.hasTriggerContent = this.triggerContentSlot.reduce((old, sl) => old || Boolean(sl.textContent.trim()), false);
```
jordanjones243, line 282: Can you add some context around this logic?

components/select/src/auro-select.js
```diff
@@ -51,7 +51,7 @@ export class AuroSelect extends LitElement {
   constructor() {
     super();
 
-    this.placeholder = 'Please select option';
+    this.placeholder = '';
```
jordanjones243, line 54: Why was this changed?

jordanjones243 (changes requested):

components/dropdown/src/auro-dropdown.js
sun-mota: I changed it to limit it to run that block only when there is no changed property. (called by `reuqestUpdate` function

components/dropdown/src/auro-dropdown.js
sun-mota: replaced to call `handleTriggerContentSlotChange`

components/select/src/auro-select.js
sun-mota: I made the change because of 3 reasons. 
1. I was told that we prefer properties to be additive.
2. placeholder is optional, not required. With that, English value as default would fail on i18n
However, I'm happy to revert back if we still want to have a default value.

jordanjones243 (changes requested):
Should be a `perf` rather than a `refactor`

components/select/src/auro-select.js
jordanjones243: Yes we should keep this for now and this should be brought up as a conversation to the team. We do not usually populate attribute with text values, but @jason-capsule42 probably knows better why the `placeholder` already has a value.

components/dropdown/src/styles/style.scss
```diff
@@ -32,7 +32,12 @@
 .label {
   font-size: var(--ds-text-body-size-xs, $ds-text-body-size-xs);
   line-height: var(--ds-text-body-size-default, $ds-text-body-size-default);
+  transition: font-size .3s cubic-bezier(.215, .61, .355, 1);
   white-space: normal;
+
+  &[hasTrigger='false'] {
+    font-size: var(--ds-text-body-size-default, $ds-text-body-size-default);
```
jordanjones243, line 39: Where are you getting these values? Is this from a design spec?

components/dropdown/src/auro-dropdown.js
```diff
@@ -236,6 +277,10 @@ export class AuroDropdown extends LitElement {
     if (changedProperties.has('mobileFullscreenBreakpoint')) {
       this.bibContent.mobileFullscreenBreakpoint = this.mobileFullscreenBreakpoint;
     }
+
+    if (changedProperties.size === 0) {
```
jordanjones243, line 281: What does this do exactly?

jordanjones243 (changes requested):

components/dropdown/src/auro-dropdown.js
sun-mota: this is the condition to know if `updated` is called by `requestUpdate`

components/dropdown/src/styles/style.scss
sun-mota: https://github.com/AlaskaAirlines/auro-formkit/issues/36
the size should be same as the placeholder font-size

components/combobox/demo/index.min.js
jason-capsule42, line 1: We need to get all these *.min.js files into the gitignore file like we did on all the other repos.

jason-capsule42 (changes requested):
I think you can have a trigger value slot and a trigger placeholder slot. 
 
When content is added to the trigger placeholder slot, we check to see if anything is in the trigger value slot. If not, we render the placeholder content.
 
When the trigger value slot changes we we check if there is content or if it was cleared out. If it was cleared out, we render the placeholder if there is any. If there is trigger value slot content we render that.
 
All of this should be doable using just the slots and applying classes to the slot containers.

components/dropdown/src/styles/style.scss
```diff
@@ -32,7 +32,12 @@
 .label {
   font-size: var(--ds-text-body-size-xs, $ds-text-body-size-xs);
   line-height: var(--ds-text-body-size-default, $ds-text-body-size-default);
+  transition: font-size .3s cubic-bezier(.215, .61, .355, 1);
```
jason-capsule42, line 35: This code is similar but not exact to what input does. I think we need handle the label styles consistently in both. There is an arguement for extracting the styles controlling the label behavior out of input and dropdown and put it in a shared SCSS file. There is probably other code we can share between the two as well like border behavior, etc.

components/dropdown/src/auro-dropdown.js
```diff
@@ -314,13 +385,13 @@ export class AuroDropdown extends LitElement {
...
           <div class="triggerContentWrapper">
-            <label class="label" id="triggerLabel">
+            <label class="label" id="triggerLabel" hasTrigger=${this.hasTriggerContent}>
               <slot name="label"></slot>
             </label>
             <div class="triggerContent">
               <slot
                 name="trigger"
-                @slotchange="${() => {this.floater.handleTriggerTabIndex(); }}"></slot>
+                @slotchange="${(event) => this.handleTriggerContentSlotChange(event)}"></slot>
```
jason-capsule42, line 394: Don't we still need to call the original function that was here before the edit too?

jason-capsule42 (changes requested):

PR #157, opened 2024-12-28
fix: use label as placeholder when there is no placeholder string set #36

### LINKED ISSUE #36: `auro-select` label doesn't move up when value is selected
## Is your feature request related to a problem? Please describe.

The current dropdown behavior is not 1:1 to the blueprint. The label should not always be shown unless there is a label AND a placeholder. Refer to the screenshot and blueprint below. 
Screen Shot 2022-05-06 at 5 02 55 PM

https://www.figma.com/file/3TQhiKjQqQOKYqaqcO1zj2/Dropdown?node-id=906%3A1298

## Describe the solution you'd like

When there is a label but no placeholder, the label should move up when there is a value selected (default state has label text in the middle and in the filled state the label moves up while the selected value takes up the middle space).

sun-mota:

Closes #36 

## Summary by Sourcery

Bug Fixes:
- Fixed an issue where the placeholder text was not displayed correctly when no placeholder string was set.

### DIFF
--- a/components/select/src/auro-select.js
+++ b/components/select/src/auro-select.js
@@ -51,7 +51,7 @@ export class AuroSelect extends LitElement {
   constructor() {
     super();
 
-    this.placeholder = 'Please select option';
+    this.placeholder = '';
     this.optionSelected = undefined;
     this.validity = undefined;
 
@@ -556,7 +556,9 @@ export class AuroSelect extends LitElement {
           </span>
           <div class="menuWrapper">
           </div>
-          <slot name="label" slot="label"></slot>
+          <span slot="label" asPlaceholder="${!this.placeholder && !this.value}">
+            <slot name="label"></slot>
+          </span>
           <span slot="helpText">
             ${!this.validity || this.validity === undefined || this.validity === 'valid'
               ? html`
--- a/components/select/demo/api.md
+++ b/components/select/demo/api.md
@@ -14,21 +14,21 @@ The auro-select element is a wrapper for auro-dropdown and auro-menu to create a
 
 ## Properties
 
-| Property                        | Attribute                       | Type      | Default                | Description                                      |
-|---------------------------------|---------------------------------|-----------|------------------------|--------------------------------------------------|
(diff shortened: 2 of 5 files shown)


components/select/src/auro-select.js
```diff
@@ -545,11 +566,13 @@ export class AuroSelect extends LitElement {
           chevron
           part="dropdown">
           <span slot="trigger" aria-haspopup="true" id="triggerFocus">
-            ${this.value ? this.displayValue : html`<span class="placeholder">${this.placeholder}</span>`}
+            ${this.value ? this.displayValue : html`<span class="placeholder">${this.placeholder || this.labelText}</span>`}
           </span>
           <div class="menuWrapper">
           </div>
-          <slot name="label" slot="label"></slot>
+          <span slot="label" class="${!this.placeholder && !this.value ? 'hidden' : ''}">
```
jason-capsule42, line 573: I am a little confused here. It looks like you are hiding the label when there is no placeholder and there is no value.

Shouldn't, like auro-input, we always show the elements label regardless of placehodler/value conditions?

components/select/src/auro-select.js
```diff
@@ -545,11 +566,13 @@ export class AuroSelect extends LitElement {
           chevron
           part="dropdown">
           <span slot="trigger" aria-haspopup="true" id="triggerFocus">
-            ${this.value ? this.displayValue : html`<span class="placeholder">${this.placeholder}</span>`}
+            ${this.value ? this.displayValue : html`<span class="placeholder">${this.placeholder || this.labelText}</span>`}
```
jason-capsule42, line 569: I think this is problematic for two reasons.

You have a span that you have identified as "placeholder" but you are also conditionally putting non-placeholder content into it. That's an anti-pattern we should avoid.

Also, the label for form elements shouldn't change as it will notably impact the experience a voice over utility relies on.

I think you are safer having two different elements that you simply hide/show the one you need based on state.

jason-capsule42: Also, I think we should be keeping slot content out of properties.

jason-capsule42 (changes requested):

sun-mota:
@sourcery-ai review

PR #383, opened 2025-03-03
fix: move children to cloned node for fullscreen slots to preserve event listener actions

sun-mota:

## Problem
any event listener to footer element does not get called because the **cloned** element of footer is displaying

## Solution
clone only the root of footer slot and move its children under cloned element so that attached listener can get property called

## Summary by Sourcery

Fixes an issue where a click event on a cloned node in the auro-counter-group component would not trigger the original node's click method. The fix adds an event listener to the cloned node that calls the original node's click method when the cloned node is clicked.

Bug Fixes:
- Fixes an issue where a click event on a cloned node in the auro-counter-group component would not trigger the original node's click method.

Chores:
- Updates the version of auro-formkit dependencies in several components.

### DIFF
--- a/components/dropdown/README.md
+++ b/components/dropdown/README.md
@@ -138,15 +138,7 @@ Running the `dev` command will open a `localhost` development server for all com
 
 To only develop a single component, use the `--filter` flag:
 
+```shell
+npx turbo dev --filter=@aurodesignsystem/auro-input
 ```
-turbo dev --filter=@aurodesignsystem/auro-input
-```
-<!-- AURO-GENERATED-CONTENT:END -->
-
-### Start formkit development environment
-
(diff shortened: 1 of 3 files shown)


components/dropdown/README.md
```diff
@@ -138,15 +138,7 @@ Running the `dev` command will open a `localhost` development server for all com
...
 To only develop a single component, use the `--filter` flag:
 
+```shell
+npx turbo dev --filter=@aurodesignsystem/auro-input
 ```
-turbo dev --filter=@aurodesignsystem/auro-input
-```
-<!-- AURO-GENERATED-CONTENT:END -->
-
-### Start formkit development environment
```
jordanjones243, line 146: Why was this last part removed? Was it intentional?

jordanjones243 (approved):

Issue #404, opened 2022-05-26
`auro-select` Type ahead to jump to items in the dropdown menu


rallyb13:
## Is your feature request related to a problem? Please describe.
We want to use auro-select for the list of countries, a very long list that takes a long time to scroll through with no option to skip/drop/pop to a point below.

## Describe the solution you'd like
Typing a letter should pop to the first entry with that letter.

## Describe alternatives you've considered
We considered using the auro-combobox in place of the select, but that had some other minor problems (no list view without typing, typing an option in  that matched a selectable option was not accepted, and behavior is strange for our use case with auto-fill).
We are currently using the native <select> element, which has typeahead support.

PR #538, opened 2025-04-29
feat: add press-key-to-jump-to-selection functionality to auro-select

StephenRiosDev:
add press-key-to-jump-to-selection functionality to auro-select

#404

## Checklist:

- [ ] My update follows the CONTRIBUTING guidelines of this project
- [ ] I have performed a self-review of my own update

-- Auro Design System Team

## Summary by Sourcery

Add press-key-to-jump-to-selection functionality to auro-select component, enabling users to quickly navigate through options by typing the first letter of their desired selection

New Features:
- Implement key press navigation for quickly finding and selecting options in the auro-select dropdown

Enhancements:
- Add ability to cycle through options that start with the same letter
- Automatically show dropdown when typing to search


components/select/src/auro-select.js
```diff
@@ -478,6 +489,45 @@ export class AuroSelect extends LitElement {
     });
   }
 
+  updateActiveOptionBasedOnKey(_key) {
```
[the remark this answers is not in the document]
StephenRiosDev: This interaction is intentional and should not cause issues even if the same key is pressed hundreds of times thanks to the use of the modulus in the implementation.

components/select/src/auro-select.js
```diff
@@ -453,13 +455,17 @@ export class AuroSelect extends LitElement {
         if (this.dropdown.isPopoverVisible) {
           this.menu.navigateOptions('down');
         }
+
+        return;
```
sun-mota, line 459: The scroll does not move to the active option with ArrowDown/ArrowUp. Is that expected behavior?

StephenRiosDev: It wasn't an existing behavior that we broke or anything, but it sounds like something that should be added. We can probably add it as part of this work if the scope creep is OK. I know Jason wants this out pretty quickly, EOD yesterday was the initial delivery date. Maybe it would be better to make a ticket to add it after this is released?

sun-mota: yes, let's make a ticket to address that.

sun-mota (approved):
approved with a known issue: scroll does not move to the active menuoption with arrow up/down keypress

sun-mota (approved):

PR #558, opened 2025-05-01
fix: not to bubble up keydown event when the input is not in bibtemplate

sun-mota:

fixing keyevent to be bubbled up correctly (only when it is in fullscreen bib) - it was coded in the opposite condition.

## Checklist:

-- Auro Design System Team

## Summary by Sourcery

Modify event handling in combobox and select components to prevent unintended key event bubbling

Bug Fixes:
- Prevent unnecessary key event propagation in combobox input handling
- Adjust event dispatching in test utilities to correctly trigger keyboard events

Enhancements:
- Refine input event management in combobox and select components

## Summary by Sourcery

Fix event handling in combobox component to correctly manage key event bubbling and prevent unintended propagation

Bug Fixes:
- Corrected key event bubbling logic in combobox input handling
- Fixed condition for preventing arrow key default behavior

Enhancements:
- Improved event management in combobox component
- Refined input event handling logic

Chores:
- Updated auro-library dependency to version 4.4.1

### DIFF
--- a/components/combobox/src/auro-combobox.js
+++ b/components/combobox/src/auro-combobox.js
@@ -610,7 +610,7 @@ export class AuroCombobox extends LitElement {
    */
   bubbleUpInputKeyEvent(event) {
     if (event.currentTarget.parentNode !== this.dropdown) {
-      if (event.key.includes("Arrow")) {
+      if (event.key === 'ArrowUp' || event.key === 'ArrowDown') {
         event.preventDefault();
       }
 
@@ -759,7 +759,7 @@ export class AuroCombobox extends LitElement {
       }));
     }
 
-    if (this.optionSelected && this.input.value !== this.optionSelected.textContent) {
+    if (this.optionSelected && this.optionSelected[0] && this.input.value !== this.optionSelected[0].textContent) {
       this.optionSelected = undefined;
       hasChange = true;
     }
--- a/components/combobox/src/auro-combobox.js
+++ b/components/combobox/src/auro-combobox.js
@@ -607,10 +607,7 @@ export class AuroCombobox extends LitElement {
    * @param {KeyboardEvent} event - The keyboard event.
    */
   bubbleUpInputKeyEvent(event) {
-    if (event.currentTarget.parentNode === this.dropdown) {
-      if (event.key === 'ArrowUp' || event.key === 'ArrowDown') {
-        event.preventDefault();
-      }
+    if (event.currentTarget.parentNode !== this.dropdown) {
       const ke = new KeyboardEvent(event.type, {
         key: event.key,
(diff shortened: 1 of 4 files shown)


components/select/src/auro-select.js
```diff
@@ -492,6 +492,11 @@ export class AuroSelect extends LitElement {
     });
   }
 
+  /**
```
jordanjones243, line 495: You should also add a function description

jordanjones243 (changes requested):
This may not be related to your PR, but I notice that the `selected` color style does not apply until you click or select the menu option twice.

Other than that, the rest of your changes look good functionally.

components/combobox/src/auro-combobox.js
```diff
@@ -607,10 +609,11 @@ export class AuroCombobox extends LitElement {
    * @param {KeyboardEvent} event - The keyboard event.
    */
   bubbleUpInputKeyEvent(event) {
-    if (event.currentTarget.parentNode === this.dropdown) {
-      if (event.key === 'ArrowUp' || event.key === 'ArrowDown') {
+    if (event.currentTarget.parentNode !== this.dropdown) {
```
jason-capsule42, line 612: I am a little confused here - the point of this code was to stop the browser from scrolling the page when you use the arrow keys but only if you are in the dropdown. Doesn't this line change negate that?

sun-mota: The point of this function `bubbleUpInputKeyEvent` was to bypass the keyboard events from input to combobox when input is moved into fullscreen bib.

components/select/src/auro-select.js
sun-mota: added :)

components/combobox/src/auro-combobox.js
sun-mota: The reason of `event.preventDefault();` in ArrowKeys is to block cursor position change in Input. Without this, The cursor moves between the letters.

components/select/src/auro-select.js
jordanjones243: The description should be above the `@private` identifier like all other functions. If you have time to fix it cool. No worries if not.

components/combobox/src/auro-combobox.js
jason-capsule42: You need to test this on a long page that would scroll and make sure that using the arrow keys when the input has focus doesn't cause the page to scroll.

sun-mota: updated to move cursor with left and right keys :)

jordanjones243 (dismissed):

jason-capsule42 (approved):

PR #569, opened 2025-05-07
Validation methodology update

StephenRiosDev:

## Checklist:

-- Auro Design System Team

## Summary by Sourcery

Update validation methodology across multiple Auro components by introducing a new 'touched' state to improve form validation behavior

New Features:
- Added 'touched' state to datepicker, input, radio group, and checkbox group components

Enhancements:
- Introduced a new 'touched' property to track user interaction across form components
- Refined validation logic to only trigger errors after user interaction

Tests:
- Updated test cases to reflect new validation behavior


StephenRiosDev:
Target branch should probably be updated to beta, but beta needs to be updated first. Targeting main for now just to get the PR out.

components/checkbox/src/auro-checkbox-group.js
```diff
@@ -39,6 +39,8 @@ export class AuroCheckboxGroup extends LitElement {
     this.required = false;
     this.horizontal = false;
     this.onDark = false;
+    this.pristine = true;
+    this.dirty = false;
```
sun-mota, line 43: JSDoc comments are needed for these variables.

components/input/src/base-input.js
```diff
@@ -422,7 +424,19 @@ export default class BaseInput extends LitElement {
...
+        type: Boolean,
+        reflect: true,
+        attribute: 'pristine'
+      },
+
+      dirty: {
+        type: Boolean,
+        reflect: true,
+        attribute: 'dirty'
+      },
```
sun-mota, line 439: 1. Please add JSDoc comments. 
2. Do these need to be public? What's the usecase of consumers set these on purpose?

components/radio/src/auro-radio-group.js
```diff
@@ -136,6 +138,14 @@ export class AuroRadioGroup extends LitElement {
...
         type: Object
+      },
+      pristine: {
+        type: Boolean,
+        reflect: true
+      },
+      dirty: {
+        type: Boolean,
+        reflect: true
       }
```
sun-mota, line 149: Same as base-input's comment. JSDoc comments needed.

packages/form-validation/src/validation.js
```diff
@@ -22,6 +22,8 @@ export default class AuroFormValidation {
   reset(elem) {
     elem.validity = undefined;
     elem.value = undefined;
+    elem.pristine = true;
+    elem.dirty = false;
```
sun-mota, line 26: if pristine and dirty will be always opposite to each other, do we really need to variables? can we just have only one?

sun-mota (changes requested):

sun-mota (changes requested):
1. Please update the PR title to represent the purpose of this PR. 
2. issue ticket should be linked.

packages/form-validation/src/validation.js
StephenRiosDev: This is an industry-standard best practice for form validation. You don't strictly need both properties, but it speaks to semantics/intent.

`if (input.dirty)` is more semantic and specific than `if (!input.pristine)` when it comes to reading code.

components/input/src/base-input.js
StephenRiosDev: The most common use case for exposing these is allowing for external state management by consumers.

sun-mota: will they actually want to do that? what if they mess this up? such as set pristine and dirty to both true or false?

StephenRiosDev:
@sourcery-ai review

components/radio/src/auro-radio-group.js
```diff
@@ -350,6 +365,8 @@ export class AuroRadioGroup extends LitElement {
    * @returns {void}
    */
   handleToggleSelected(event) {
+    this.touched = true;
```
sun-mota, line 368: input and checkbox's touched become `true` on  `focusin` but radio does on `toggleSelected`. 
Should they be unified?

sun-mota: Adding `focusin` event handler is not a huge thing IMO. I'd rather not leave it as tech debt.

packages/form-validation/src/validation.js
```diff
@@ -229,11 +230,17 @@ export default class AuroFormValidation {
     this.getAuroInputs(elem);
 
     // Validate only if noValidate is not true and the input does not have focus
-    const validationShouldRun = force || (!elem.contains(document.activeElement) && elem.value !== undefined) || elem.validateOnInput;
+    let validationShouldRun =
+      force ||
+      (!elem.contains(document.activeElement) &&
+        (elem.touched ||
+          (!elem.touched && typeof elem.value !== "undefined"))) ||
```
jordanjones243, line 237: Why was the addition of `typeof` needed?

jordanjones243 (changes requested):
- `radio` required validation is NOT working
- Not seeing help text for `patternMistmatch` state in input (validate on input example)
- `select` required validation is NOT working
- Datepicker range support has interesting UI: When having two values, deleting the second one via the `x` icon puts the datepicker in a `valueMissing` state without shifting focus from input. Then, after selecting a date with the calendar, we still see the error icon until the input loses focus. Same can actually be said with non-range datepicker.

components/radio/src/auro-radio-group.js
StephenRiosDev: We can absolutely copy the lines of code that handle this in checkbox over to radio and make the necessary adjustments to get it to work, but I have some concerns with this approach, especially since the naming conventions for the events aren't consistent and the code appears rather fragile, and we'd be duplicating a lot of code that should probably be standardized instead.

I'm happy to take this work up before moving on to my next sprint task if you're worried about it sitting, it won't be an issue for me to get it resolved quickly before moving on.

Let's connect in the morning and see if we can come to a solution that works for us to get this work out the door and still leave ourselves room for any necessary improvements we need to make in the future.



For reference, this is the code that would need to be copied and modified:

```javascript
// must declare this function as a variable to correctly pass the reference to the removeEventListener
const checkFocusWithin = function(evt) {
  if (document.auroCheckboxGroupActive && !document.auroCheckboxGroupActive.contains(evt.target)) {
    // if focus has left the group, cleanup and validate
    document.auroCheckboxGroupActive.focusWithin = false;
    window.removeEventListener('focusin', checkFocusWithin);
    document.removeEventListener('click', checkFocusWithin);
    // execute the validation
    document.auroCheckboxGroupActive.validation.validate(document.auroCheckboxGroupActive);
  }
};

this.addEventListener('auroCheckbox-focusin', () => {
  this.touched = true;

  if (!this.value) {
    this.value = [];
  }

  // handle click outside the group
  if (!this.focusWithin) {
    document.addEventListener('click', checkFocusWithin);
  }

  this.focusWithin = true;
});

this.addEventListener('auroCheckbox-focusout', () => {
  document.auroCheckboxGroupActive = this;

  // Only add the focusWithin check event listener once as you move focus through the options
  if (this.focusWithin) {
    window.addEventListener('focusin', checkFocusWithin);
  } else {
    this.focusWithin = true;
  }
});
```

packages/form-validation/src/validation.js
StephenRiosDev: `typeof value !== 'undefined'` is a safeguard against reference errors for undeclared variables and properties, kind of like how optional chaining helps us with reference errors but for `undefined` values specifically.

StephenRiosDev:
> * `radio` required validation is NOT working

I re-tested this, it seems to be working as I would expect and how it currently works for me on the docs site and main. Interested to see what it does for you.
 
> * Not seeing help text for `patternMistmatch` state in input (validate on input example)

This seems to be working for me, but only on blur, which is the same way this works on the docs site and on the current `main` branch, which I do agree seems odd. I totally expected it to validate as I typed.

> * `select` required validation is NOT working

This got skipped somehow? Good catch, thank you. It's been added.

> * Datepicker range support has interesting UI: When having two values, deleting the second one via the `x` icon puts the datepicker in a `valueMissing` state without shifting focus from input. Then, after selecting a date with the calendar, we still see the error icon until the input loses focus. Same can actually be said with non-range datepicker.

Yah this is definitely strange behavior, but it's doing this on the docs site and main as well.

components/checkbox/src/auro-checkbox-group.js
```diff
@@ -307,6 +317,7 @@ export class AuroCheckboxGroup extends LitElement {
       checkbox.reset();
     });
 
+    this.touched = false;
```
jordanjones243, line 320: You should be setting `this.touched = false` in all component's `reset()` method

jordanjones243 (changes requested):

sun-mota (changes requested):
image
1. no error message showing on required select.


2. required checkbox group validation does not run on blur.

sun-mota (approved):

Issue #589, opened 2025-05-12
a11y: `auro-input` contains duplicated id under shadowDom and aria-label is set wrong on `auro-button` inside.


sun-mota:
### What browser client(s) are you seeing the issue on?

- [x] Firefox
- [x] Chrome
- [x] Safari
- [x] Microsoft Edge

### What devices platform(s) are you seeing the issue on?

- [x] Desktop
- [x] Mobile

### Steps to reproduce

1. `auro-input` and it's input element under shadowRoot have the exact same id. And id should be unique. 

![Image](https://github.com/user-attachments/assets/1b5410ca-eb9a-4b9a-a209-a570c6b548eb)

2. `aria-label` attribute for button is at wrong place (on auro-button, not on actual button node)

![Image](https://github.com/user-attachments/assets/e20a44c6-7485-4e6e-b14a-ddaaf2837d93)

### Reproducible on the Auro doc site

- [ ] I am able to reproduce this issue on the [Auro doc site](https://auro.alaskaair.com).

PR #590, opened 2025-05-12
fix: set unique id and arialabel correctly in input/radio/checkbox #589

sun-mota:

1. make input node under shadowRoot to have a unique id (not the dupId from its parent)
2. set `arialabel` on auro-button instead of `aria-label`

## Checklist:

-- Auro Design System Team

## Summary by Sourcery

Fix input element id handling and accessibility attributes in `auro-input`. Introduce a dedicated `inputId` for the input and update the label’s `for` and the input’s `id` to use it, correct the clear-button attribute name, and adjust tests to select the input element accordingly.

Bug Fixes:
- Generate and use a unique `inputId` for the input element and link the label’s `for` attribute to it instead of using the host’s `id`.
- Correct the clear-button attribute from `aria-label` to `arialabel` to match expected markup.

Tests:
- Update tests to query the `<input>` element directly instead of selecting by the host `id`.

### DIFF
--- a/components/checkbox/docs/api.md
+++ b/components/checkbox/docs/api.md
@@ -51,7 +51,7 @@ Custom element for the purpose of allowing users to select one or more options o
 | `checked`  | `checked`  | `boolean` | false   | If set to true, the checkbox will be filled with a checkmark. |
 | `disabled` | `disabled` | `boolean` | false   | If set to true, the checkbox will not be clickable. |
 | `error`    | `error`    | `boolean` | false   | If set to true, the checkbox will be displayed with an error state. |
-| `id`       | `id`       | `string`  |         | Sets the individual `id` per element.            |
+| `id`       | `id`       | `string`  |         | The id global attribute defines an identifier (ID) which must be unique in the whole document. |
 | `name`     | `name`     | `string`  |         | Accepts any string and is used to identify related checkboxes when submitting form data. |
 | `onDark`   | `onDark`   | `boolean` | false   | Sets onDark styles for component.                |
 | `value`    | `value`    | `string`  |         | Sets the element's input value. Must be unique within an auro-checkbox-group element. |
--- a/components/checkbox/src/auro-checkbox.js
+++ b/components/checkbox/src/auro-checkbox.js
@@ -18,6 +18,9 @@ import AuroLibraryRuntimeUtils from '@aurodesignsystem/auro-library/scripts/util
 /**
  * Custom element for the purpose of allowing users to select one or more options of a limited number of choices.
  *
+ * @prop {string} id - The id global attribute defines an identifier (ID) which must be unique in the whole document.
+ * @attr id
+ *
  * @csspart checkbox - apply css to a specific checkbox.
  * @csspart checkbox-input - apply css to a specific checkbox's input.
  * @csspart checkbox-label - apply css to a specific checkbox's label.
@@ -85,11 +88,6 @@ export class AuroCheckbox extends LitElement {
       },
 
       /**
-       * Sets the individual `id` per element.
-       */
-      id: { type: String },
-
-      /**
        * Accepts any string and is used to identify related checkboxes when submitting form data.
        */
       name: { type: String },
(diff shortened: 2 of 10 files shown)


blackfalcon (changes requested):

components/checkbox/src/auro-checkbox.js
```diff
@@ -243,13 +250,13 @@ export class AuroCheckbox extends LitElement {
           @input="${this.handleInput}"
           ?disabled="${this.disabled}"
           .checked="${this.checked}"
-          id="${ifDefined(this.id)}"
+          id="${this.inputId}"
```
blackfalcon, line 253: Okay, just so I understand this correctly, the manual `id` was removed from the API, and a custom-generated ID is always generated for the component.

components/checkbox/src/auro-checkbox.js
```diff
@@ -209,6 +214,8 @@ export class AuroCheckbox extends LitElement {
     // Add the tag name as an attribute if it is different than the component name
     this.runtimeUtils.handleComponentTagRename(this, 'auro-checkbox');
 
+    this.inputId = this.id ? `${this.id}-input` : window.crypto.randomUUID();
```
blackfalcon, line 217: I'd put a depreciation comment here stating that this code is taking into account any legacy `id` attributes and mutating the value. I would assume that now that `id` has been removed from the public API, future use will begin to ignore that capability. 

In future iterations leading up to a MAJOR release, this code can then simply ignore any `id` attribute in favor of the randomUUID.

components/input/src/base-input.js
```diff
@@ -473,6 +476,7 @@ export default class BaseInput extends LitElement {
     if (this.tagName.toLowerCase() !== 'auro-input') {
       this.setAttribute('auro-input', true);
     }
+    this.inputId = this.id ? `${this.id}-input` : window.crypto.randomUUID();
```
blackfalcon, line 479: Same feedback about the deprecation comment.

components/radio/src/auro-radio.js
```diff
@@ -231,6 +240,8 @@ export class AuroRadio extends LitElement {
 
     this.input = this.shadowRoot.querySelector('input');
 
+    this.inputId = this.id ? `${this.id}-input` : window.crypto.randomUUID();
```
blackfalcon, line 243: Time for a hat trick.

blackfalcon (changes requested):
Outside of the requested comments, this is ready to ship. Good updates.

components/checkbox/src/auro-checkbox.js
sun-mota: `id` is not removed. It's a attr/prop in `HTMLElement`. So it won't get ignored.

blackfalcon (changes requested):
This change constitutes a breakign change and cannot be shipped as a `fix:`. I would recommend finding a solution that allows for deprecation of the legacy API while allowing for backwards compatibility.

blackfalcon (approved):

components/checkbox/docs/api.md
```diff
@@ -51,7 +51,7 @@ Custom element for the purpose of allowing users to select one or more options o
 | `checked`  | `checked`  | `boolean` | false   | If set to true, the checkbox will be filled with a checkmark. |
 | `disabled` | `disabled` | `boolean` | false   | If set to true, the checkbox will not be clickable. |
 | `error`    | `error`    | `boolean` | false   | If set to true, the checkbox will be displayed with an error state. |
-| `id`       | `id`       | `string`  |         | Sets the individual `id` per element.            |
+| `id`       | `id`       | `string`  |         | The id global attribute defines an identifier (ID) which must be unique in the whole document. |
```
rmenner, line 54: I would like to add more context to the descriptions for id.
(maybe something like this)
If no id is provided, a UUID will be generated and used as the `id`. The `id` of the input element within the component will be this value with the string `-input` appended.

sun-mota: UUID will be generated not for `id` but for `inputId`(which is private)
The description for `id` is from mdn page. (https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/id) as Jason's suggestion.

rmenner: Didn't think about it that way.

rmenner (approved):

PR #607, opened 2025-05-15
fix: update check for hasValue to accommodate array value types

StephenRiosDev:

## Checklist:

-- Auro Design System Team

## Summary by Sourcery

Bug Fixes:
- Prevent arrays without elements or with an empty first element from counting as having a value in the validation logic.

### DIFF
--- a/packages/form-validation/src/validation.js
+++ b/packages/form-validation/src/validation.js
@@ -260,14 +260,11 @@ export default class AuroFormValidation {
       }
 
       // Check array value types for having a value
-      if (typeof elem.value === "object") {
-        if (!elem.value.length) {
-          hasValue = false;
-        }
-
-        if (typeof elem.value[0] !== "string" || !elem.value[0].length) {
-          hasValue = false;
-        }
+      if (Array.isArray(elem.value)) {
+        hasValue = Boolean(
+          elem.value.length > 0 &&
+          elem.value.some((value) => typeof value === "string" && value.length > 0)
+        )
       }
 
       // If there is a second input in the elem and that value is undefined or an empty string set hasValue to false;
--- a/packages/form-validation/src/validation.js
+++ b/packages/form-validation/src/validation.js
@@ -252,7 +252,23 @@ export default class AuroFormValidation {
        * The validityState definitions are located at https://developer.mozilla.org/en-US/docs/Web/API/ValidityState.
        */
 
-      let hasValue = elem.value && elem.value.length > 0;
+      let hasValue = false;
+
+      // Check string for having a value
+      if (typeof elem.value === "string") {
+        hasValue = elem.value && elem.value.length > 0;
+      }
+
+      // Check array value types for having a value
+      if (typeof elem.value === "object") {
+        if (!elem.value.length) {
+          hasValue = false;
+        }
+
+        if (typeof elem.value[0] !== "string" || !elem.value[0].length) {
+          hasValue = false;
+        }
+      }
 
       // If there is a second input in the elem and that value is undefined or an empty string set hasValue to false;
       if (this.auroInputElements?.length === 2) {


packages/form-validation/src/validation.js
```diff
@@ -252,7 +252,20 @@ export default class AuroFormValidation {
...
+      if (typeof elem.value === "string") {
+        hasValue = elem.value && elem.value.length > 0;
+      }
+
+      // Check array value types for having a value
+      if (Array.isArray(elem.value)) {
+        hasValue = Boolean(
+          elem.value.length > 0 &&
+          elem.value.some((value) => typeof value === "string" && value.length > 0)
+        )
```
DukeFerdinand, line 267: doesn't `some` already return a boolean?

sun-mota, line 267:
```suggestion
        hasValue = elem.value.some((value) => typeof value === "string" && !!value);
```
This is sufficient as false will get returned when the array has no item.

packages/form-validation/src/validation.js
```diff
@@ -252,7 +252,20 @@ export default class AuroFormValidation {
        * The validityState definitions are located at https://developer.mozilla.org/en-US/docs/Web/API/ValidityState.
        */
 
-      let hasValue = elem.value && elem.value.length > 0;
+      let hasValue = false;
+
+      // Check string for having a value
+      if (typeof elem.value === "string") {
+        hasValue = elem.value && elem.value.length > 0;
```
sun-mota, line 259:
```suggestion
        hasValue = !!elem.value;
```
nit: can be simplified like this.

sun-mota (changes requested):

packages/form-validation/src/validation.js
StephenRiosDev: > doesn't `some` already return a boolean? if so we don't need the Boolean() wrapper :D

The Boolean wrapper lets us extend this check with extra checks if we want to, we already have two. I'd say it was unnecessary if we just had: 

```javascript
hasValue = Boolean(
  elem.value.some((value) => typeof value === "string" && value.length > 0)
)
```

But there's two checks, and this lets us extended with more if we want:

```javascript
hasValue = Boolean(
  elem.value.length > 0 &&
  elem.value.some((value) => typeof value === "string" && value.length > 0) &&
  elem.value.filter((value) => value.length > 3)?.length // at least one value is longer than 3
)
```

packages/form-validation/src/validation.js
StephenRiosDev: I think keeping this more explicit to it's purpose (checking that the length is longer than a minimum threshold) is good here.

If ESLint would let me, I'd prefer to have it written as:

```javascript
hasValue = elem.value?.length > 0;
```

ESlint also won't let us use the double-not operator anyways, which is sad, because I'm a big fan.

packages/form-validation/src/validation.js
sun-mota: I understand open/closed principal but how is this not extendable?

we can extend like this.
```
hasValue = 
  elem.value.some((value) => typeof value === "string" && value.length > 0) &&
  elem.value.filter((value) => value.length > 3)?.length // at least one value is longer than 3
```

StephenRiosDev: I'm not sure I'm understanding what your argument is.

Why shouldn't we have an explicit and adjustable threshold check?

```javascript
Boolean(
  elem.value.length > 0 && // This line lets us make sure there's N values, which could be adjustable later without modification to the code
  elem.value.some((value) => typeof value === "string" && value.length > 0) // Other checks run as normal
)
```
If later we wanted to make a threshold value for menu's required so that it required at minimum two selections, we could just change that line to something like:
```javascript
Boolean(
  elem.value.length > elem.minSelectedValues &&
  elem.value.every((value) => typeof value === "string" && value.length > 0)
)
```
This helps make future changes easier, commits more readable, etc.

DukeFerdinand: I'm a fan of this! resolving this convo since we talked about it off github :)

sun-mota (approved):

PR #1112, opened 2025-08-19
fix(datepicker): clicking bib no longer breaks on safari/firefox

DukeFerdinand:

Safari and Firefox were both having issues where clicking on the dropdown bib would close the bib, no matter what. Turns out this is an issue with BOTH browsers where they fire `focusout` when they shouldn't be. Our code was rightfully checking to see if the event target was the datepicker itself, but because they don't follow what Chrome does (and contain/not bubble up) we were getting a false positive and closing the datepicker. 

We also don't _need_ to update the datepicker dropdown manually as the bib takes care of closing itself when it's blurred.

## Checklist:

-- Auro Design System Team

## Summary by Sourcery

Fix datepicker closing unexpectedly on Safari and Firefox when interacting with the bib dropdown handle by removing an unnecessary hide call and refining focus logic to ignore clicks inside the bib content.

Bug Fixes:
- Remove explicit dropdown.hide() call on focusout to prevent premature closure.
- Add a check to skip focus-handling when clicks originate within the bibContent in Safari/Firefox.

### DIFF
--- a/components/datepicker/src/auro-datepicker.js
+++ b/components/datepicker/src/auro-datepicker.js
@@ -806,7 +806,6 @@ export class AuroDatePicker extends AuroElement {
 
       if (!this.contains(document.activeElement)) {
         this.validate();
-        this.dropdown.hide();
       }
     });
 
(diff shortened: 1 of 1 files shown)


components/datepicker/src/auro-datepicker.js
```diff
@@ -1246,7 +1245,7 @@ export class AuroDatePicker extends AuroElement {
     const targetIsInput = initTarget.tagName === 'INPUT';
     const isFocusAlreadyOnInput = this.inputList.includes(this.shadowRoot.activeElement);
 
-    if (layoutRequiresHandling && !targetIsInput && !isFocusAlreadyOnInput) {
+    if (layoutRequiresHandling && !targetIsInput && !isFocusAlreadyOnInput && !event.composedPath().includes(this.dropdown.bibContent)) {
```
jason-capsule42, line 1248: I suspect event.composedPath doesn't work in NodeJS environment and will break the Borealis build. Probably need a try/catch on this.

DukeFerdinand: we talked about this on a call, determined it should be fine as is since it's 1. not executed in node, and 2. already being used by Borealis as of 1 month ago

rmenner (approved):

Issue #1214, opened 2025-11-07
auro-input: updating `placeholder` attribute does not get reflected visually right away


sun-mota:
### What browser client(s) are you seeing the issue on?

- [ ] Firefox
- [x] Chrome
- [ ] Safari
- [ ] Microsoft Edge

### What devices platform(s) are you seeing the issue on?

- [x] Desktop
- [ ] Mobile

### Steps to reproduce

on inspector, update `placeholder` attribute
Check if input updates it

### Expected behavior

placeholder element should update the text immediately

### Reproducible on the Auro doc site

- [x] I am able to reproduce this issue on the [Auro doc site](https://auro.alaskaair.com).

PR #1215, opened 2025-11-07
fix: update `placeholderStr` when `placeholder` attr gets changed #1214

sun-mota:

Fix to show updated `placeholder` attr value immediately when it's changed. 

## Checklist:

-- Auro Design System Team

## Summary by Sourcery

Bug Fixes:
- Invoke getPlaceholder in updated() when the placeholder property changes to immediately reflect the new placeholder value

### DIFF
--- a/components/input/src/base-input.js
+++ b/components/input/src/base-input.js
@@ -93,7 +93,6 @@ export default class BaseInput extends AuroElement {
     this.validationCCLength = undefined;
     this.hasValue = false;
     this.label = 'Input label is undefined';
-    this.placeholderStr = '';
 
     this.allowedInputTypes = [
       "text",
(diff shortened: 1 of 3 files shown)


components/input/src/base-input.js
StephenRiosDev: If changing `hasValue` should have the effect of causing an update, it should be defined as a reactive property so that we don't have to call `requestUpdate` when it changes.

StephenRiosDev: Since changing `placeholderStr` should cause a re-render since the placeholder string was update, `placeholderStr` should be defined as a reactive property so we don't have to manually call `requestUpdate`.

StephenRiosDev (dismissed):

PR #1269, opened 2026-01-05
fix: enhance checkbox a11y & add support for VoiceOver

chrisfalaska:

## The Problem

When a VoiceOver user double-tapped the `checkbox`, the tap event was hitting the host element, but the accessibility information was coming from the native `input` buried inside the shadow DOM. This disconnect caused the double-tap to fail to properly toggle the checkbox.

- VoiceOver was trying to interact with what it "saw" (potentially the host or label)
- But the actual interactive element (the native input) was hidden and not receiving the interaction correctly

**Error on Auro DocSite:**

https://github.com/user-attachments/assets/ac665fae-58c9-47b5-804a-e9db1aea188a

**Error on Production:**

https://github.com/user-attachments/assets/dde5ab28-0013-4496-a707-1b62739dc3d3

## The Fix

The initial suggestion in the ticket was to add `aria-hidden="true"` to the container. However, this was not sufficient because it only hides the element from screen readers; it does not remove the element from the interaction model.

With only `aria-hidden="true"` on the container:

- The native input was still there and still the accessible checkbox
- VoiceOver could still interact with it by doubr-tapping
- But this created an accessibility violation because `aria-hidden="true"` was on a container with a focusable element inside
- Additionally, checkboxes did not audibly announce when they were checked and unchecked as a point of confirmation

Instead, a more holistic change was made: the host element itself was made the checkbox from an accessibility perspective.

- `role="checkbox"` tells assistive technology that this element *is* a checkbox
- `aria-checked` communicates the state directly on the element the user is interacting with
- `tabindex="0"` makes the host focusable
- The native input becomes purely an implementation detail (`display: none`)
- Checkboxes audibly announce their state (checked or unchecked)

This was modeled after the pattern used in `auro-radio`: the native input exists only for form submission semantics; accessibility is handled entirely through ARIA on the host.

- Added `role="checkbox"` on the host element (as a reflected property)
- Removed `delegatesFocus: true` from the shadow root options (the input was changed to `display: none`, therefore no longer needed)
- Changed the input class from `util_displayHiddenVisually` to `util_displayHidden` (uses `display: none` to fully remove it from the accessibility tree), matching `auro-radio`
- Added `aria-hidden="true"` and `tabindex="-1"` on the input element
- Added an `updated()` method that sets `aria-checked` and `aria-disabled` on the host when properties change, so state changes are announced audibly when checked/unchecked
- Set `tabindex="0"` on the host in `firstUpdated()` for keyboard accessibility
- Removed the `label` association -- no longer pointing to the hidden `input`
  - The host element is now the sole `checkbox` in the accessibility tree
  - Slot content provides the accessible name for the host element with `role="checkbox"`

## Testing

- Validated fix using VoiceOver on iPhone 15 in Safari & Chrome
- Accessibility Inspector on MacOS
- LevelAccess code scan
- QA reviewed with @blackfalcon 

### Fix in Situ

**Corrected Behavior:**

https://github.com/user-attachments/assets/d40deadb-c909-4219-a47f-c34229503115

**Accessible tree validated:**

image

## Checklist:

-- Auro Design System Team

## Summary by Sourcery

Enhancements:
- Hide the internal checkbox input container from the accessibility tree via aria-hidden to reduce confusion for assistive technologies.

## Summary by Sourcery

Improve auro-checkbox accessibility by making the host element the primary checkbox control and aligning behavior with screen readers like VoiceOver.

New Features:
- Expose the checkbox host element as the interactive control via ARIA attributes, including role="checkbox", aria-checked, aria-disabled, and tabindex for keyboard access.

Bug Fixes:
- Resolve VoiceOver interaction issues where double-taps targeted the host while accessibility semantics were tied to a hidden native input.

Enhancements:
- Derive the accessible name for the checkbox from slot content on the host instead of associating a label to the hidden input.
- Hide the internal input from the accessibility tree and interaction model while retaining it for form submission semantics.
- Remove shadow DOM focus delegation and update styles to manage focus outline on the host only.

Tests:
- Add tests to verify role, aria-checked, aria-disabled, and tabindex behavior on the checkbox host element.

### DIFF
--- a/components/checkbox/src/auro-checkbox.js
+++ b/components/checkbox/src/auro-checkbox.js
@@ -41,20 +41,14 @@ export class AuroCheckbox extends LitElement {
     this.error = false;
     this.onDark = false;
     this.touched = false;
+    this.tabIndex = 0;
+    this.ariaChecked = 'false';
+    this.role = 'checkbox';
 
     /**
      * @private
      */
     this.runtimeUtils = new AuroLibraryRuntimeUtils();
-
-    /**
-     * @private
-     * @property {boolean} delegatesFocus - Whether the shadow root delegates focus.
-     */
-    this.constructor.shadowRootOptions = {
-      ...LitElement.shadowRootOptions,
-      delegatesFocus: true,
-    };
   }
 
   static get styles() {
@@ -144,7 +138,48 @@ export class AuroCheckbox extends LitElement {
         type: String,
         reflect: false,
         attribute: false
-      }
+      },
+
+      /**
+       * The tabindex attribute for the checkbox.
+       * @private
+       */
+      tabIndex: {
+        type: Number,
+        reflect: true,
+        attribute: 'tabindex'
+      },
+
+      /**
+       * The aria-checked attribute for the checkbox.
+       * @private
+       */
+      ariaChecked: {
+        type: String,
+        reflect: true,
+        attribute: 'aria-checked'
+      },
+
+      /**
+       * The aria-disabled attribute for the checkbox.
+       * @private
+       */
+      ariaDisabled: {
+        type: String,
+        reflect: true,
+        attribute: 'aria-disabled'
+      },
+
+      /**
+       * The ARIA role for the element. Must remain 'checkbox' for screen readers
+       * to correctly identify this as a checkbox control.
+       * @private
+       */
+      role: {
+        type: String,
+        reflect: true
+      },
+
     };
   }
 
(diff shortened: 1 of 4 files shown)


blackfalcon (approved):

chrisfalaska:
During review, a question was raised about whether `aria-label` was necessary.

The host element itself is the checkbox control, with:

```
role="checkbox"
tabindex="0"
aria-checked
```

Because of this, it requires an accessible name.

The `aria-label`, derived from the slot content, provides that name. The slot content is visually displayed as the label, but since it resides in the light DOM and the host is the actual focusable element with the `checkbox` role, screen readers need the `aria-label` on the host to announce the name.

Without it, a screen reader would announce something like "checkbox, not checked," with no indication of what the checkbox is for.

jason-capsule42 (approved):

Issue #1272, opened 2026-01-06
Combobox: `error` attribute does not present any change in UI


jordanjones243:
### What browser client(s) are you seeing the issue on?

- [ ] Firefox
- [x] Chrome
- [ ] Safari
- [ ] Microsoft Edge

### What devices platform(s) are you seeing the issue on?

- [x] Desktop
- [ ] Mobile

### Steps to reproduce

<img width="578" height="363" alt="Image" src="https://github.com/user-attachments/assets/9289fad7-27d7-4248-bc28-8fcf7051aa32" />

### Expected behavior

There should be an error border when the `error` attribute is present.

### Which version of auro-formkit do you have installed?

5.9.1

### Reproducible on the Auro doc site

- [x] I am able to reproduce this issue on the [Auro doc site](https://auro.alaskaair.com).

### Exit criteria (To be determined by the Auro team)

Error border exists in UI when `error` attribute is present

jordanjones243:
I think the fix should be adding `.error=${this.error}` to the `auro-input` HTML template inside the `auro-combobox.js` file.

PR #1307, opened 2026-01-26
fix(combobox): error attribute now correctly displays error UI #1272

chrisfalaska:

Fixed a bug where the `error` attribute on `auro-combobox` did not display any visual error state in the UI.                                                                                 

This was because the validation logic in `validation.js` was overwriting the `customError` validity state with the child input's validity.

  - Added `elem.validity !== "customError"` check to prevent overwriting the validity when the `error` attribute is set.
  - Fixed error attribute propagation to properly handle both setting and removing the error attribute
  - Added test to verify error attribute correctly sets `customError` validity state  

Additionally, I removed the `async`/`await` from 931c296 that @jordanjones243 flagged. These never worked as intended because `setAttribute()` is synchronous and doesn't return a Promise, so the await had no effect.

All tests for Datepicker & Combobox pass.

## Primary Bug

In `validation.js`:

**Before:**

```
if (this.auroInputElements?.length > 0 && elem.validity !== "valueMissing")
```

**After:**

```
if (this.auroInputElements?.length > 0 && elem.validity !== "valueMissing" && elem.validity !== "customError")
```

This prevents the validation logic from overwriting `customError` (set when error attribute exists) with the child input's validity state.

## Secondary Cleanup Issue

In `auro-combobox.js`:

Before:

```
await this.input.setAttribute('error', this.getAttribute('error'));
```

After

```
if (this.hasAttribute('error')) {
  this.input.setAttribute('error', this.getAttribute('error'));
} else {
   this.input.removeAttribute('error');
}
```

The problem was when the error attribute was removed from combobox, `this.getAttribute('error')` returned `null`. Then `setAttribute('error', null)` set the attribute to the literal string "null" instead of removing it.

The new code explicitly handles both cases:

- Error is set: propagate the error message to the child input
- Error is removed: actually remove the attribute from the child input (not set it to "null")

## TLDR

- Primary bug (error UI not showing): Fixed in `validation.js` by preserving `customError` validity
- Secondary cleanup (proper attribute removal): Fixed with the `hasAttribute`/`removeAttribute` in `auro-combobox.js`

## Testing

**Before:**

Screenshot 2026-01-26 at 3 18 46 PM

**After:**

Screenshot 2026-01-26 at 3 18 57 PM

Screenshot 2026-01-26 at 3 19 06 PM

## Checklist:

-- Auro Design System Team

## Summary by Sourcery

Fix error state handling for auro-combobox and related components to correctly reflect the error attribute and preserve custom validation states.

Bug Fixes:
- Ensure auro-combobox displays an error state and customError validity when the error attribute is set and clears it when removed.
- Prevent form validation from overwriting an element's customError validity state when syncing validity from child inputs.
- Correct error attribute propagation on auro-combobox and auro-datepicker inputs, including proper removal of the attribute when cleared.

Enhancements:
- Simplify lifecycle methods by removing unnecessary async/await usage from updated() in auro-combobox and auro-datepicker.

Tests:
- Add a regression test verifying that setting and removing the error attribute on auro-combobox correctly updates its validity and error message.

### DIFF
--- a/components/combobox/test/auro-combobox.test.js
+++ b/components/combobox/test/auro-combobox.test.js
@@ -563,8 +563,8 @@ function runFulltest(mobileview) {
     el.validate(true);
     await elementUpdated(el);
 
-    // Should clear the customError state (validity should not be customError anymore)
-    await expect(el.getAttribute('validity')).to.not.equal('customError');
+    // Should clear the customError state and return to valid
+    await expect(el.getAttribute('validity')).to.be.equal('valid');
   });
 }
 


jordanjones243 (changes requested):
Code changes and functionality look good, but needs to be rebased to `dev`. Will approve once addressed.

PR #1474, opened 2026-05-12
Remove disabled form element submissions

jordanjones243:

https://itsals.visualstudio.com/E_Retain_Content/_workitems/edit/1541873

> **Branch target:** This PR targets `breaking/formkit-v6` and is intended as an intentional v6 breaker. It introduces consumer-visible behavior changes to `form.value`, `form.validity`, `form.isInitialState`, submit/reset semantics, and Enter-key handling. See the **Behavioral decisions (round 2 review)** section below for a per-decision rationale and sign-off list.

## Review Checklist:

<details>
<summary>
RC Checklist
</summary>

### Browsers

Android

- [ ] Chrome
- [ ] Firefox

 iOS

- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Desktop

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### Scenarios

- [ ] Next React
- [ ] SvelteKit

</details><br>

-- Auro Design System Team

## Summary by Sourcery

Align auro-form behavior with HTML spec by excluding disabled controls from submission, validation, and initial-state logic, and document and demo this behavior.

Bug Fixes:
- Exclude disabled form controls from auro-form.value and from the submit event payload.
- Prevent disabled required fields from marking the form invalid or blocking submission.
- Ensure Enter key on disabled controls does not trigger form submission.

Enhancements:
- Track disabled attribute changes on child form elements so form state and button enablement update reactively.
- Update column layout example markup and styles to better demonstrate grid-based form layouts.
- Add a documentation section explaining how disabled fields interact with auro-form state, including a live example wired to show submitted values.

Tests:
- Add test coverage verifying disabled fields are omitted from form value and submission, do not affect validity, and can be re-included when re-enabled.

### Deferred to follow-up

- **O(n²) lookup pattern in `_isNameDisabled`**: cost is `_elements.find()` per `formState` key on every read of `value` / `validity` / `isInitialState`. Acceptable for forms with < ~50 fields (covers all current consumers). Documented in a code comment near the helper. The proper fix is to store `disabled` on each `formState` entry and update it from the attribute observer; that would ripple through the `FormStateMember` typedef and existing `deep.equal` assertions in tests, which is more churn than this PR should carry.
- **`<fieldset disabled>` inheritance**: would require walking ancestors in `_isDisabled` (or pre-computing it on the observer pass). No precedent in the codebase and not requested by the ticket.
- **`auro-counter-group` lacks a `disabled` property**: out of scope per the original plan; tracked as a separate concern.

## Behavioral decisions (round 2 review)

This section records the behavioral decisions made in this PR and the rationale behind each. Reviewers can use it to sign off per decision, or push back where they see things differently. Each section ends with the specific question I'd like the reviewer to answer.

The PR's core intent is to bring `auro-form` in line with native HTML form semantics around disabled controls. In the course of getting there, a few adjacent design questions surfaced — each is called out below.

---

### Decision 1 — Disabled controls are excluded from value, validity, and submission

**Decision.** A control with the `disabled` attribute is skipped by:
- `form.value` (its `name` is omitted from the object)
- `form.validity` (it can't make the form invalid)
- The `submit` event's `detail.value` (omitted)
- `form.submit()` force-revalidation (`validate(true)` is not called on it)
- Enter-key submission when the Enter originated from a disabled control (returns early)

**Why.** Native `<form>` semantics: disabled controls don't submit, aren't validated, and don't participate in constraint validation. `auro-form` previously treated disabled like any other field, which created two concrete user-facing problems:
1. A disabled + required field would block submission for a form the user had otherwise completed.
2. A disabled field carrying a value (e.g., a read-only account number) would land in the submit payload.

**What changes for consumers.** Code reading `form.value.disabledFieldName` now gets `undefined`. Code that relied on disabled+required to block submission needs a different pattern — the documentation now recommends `<input type="hidden">` for "include this value but don't show or validate it."

**Trade-off / pushback room.** This is a meaningful API change. Consumers using `disabled` to carry values through to submission will need to migrate. Justified by the platform-alignment goal.

> **Reviewer ask:** Sign off that aligning with native HTML disabled semantics is the right primary goal of this PR?

---

### Decision 2 — `isInitialState` adopts HTML's `dirtyValueFlag` model

**Decision.** A field taints the form if its current value differs from its captured initial value, OR if its validity is in an error state (not `null` and not `'valid'`). The form captures each field's initial value once on first sight and preserves it across `initializeState` cycles (slot change, rename, reset). Disabled state is intentionally not short-circuited — disabling a field that the user has already edited does NOT clear its dirty flag.

**Why.** The original PR added a `_isNameDisabled` skip to `_setInitialState` so that disabled fields didn't taint. Jason flagged the side effect: if a user edits a field and then JS disables that field, the form silently flips back to "initial," the Reset button re-disables itself, and the dirty state is lost. That was the bug. The fix is HTML's existing model — the dirty-value flag is per-field and isn't cleared by disabling the control.

**What changes for consumers.**

| Scenario | `dev` | This PR |
|---|---|---|
| Empty form, untouched | `isInitialState = true` | `isInitialState = true` |
| `<auro-input value="hello">`, no edit | `false` (any value tainted) | `true` (current === initial) |
| `<auro-input value="hello" disabled>`, no edit | `false` | `true` |
| User types | `false` | `false` |
| User edits → JS disables the field | `true` (lost dirty state) | `false` (still dirty) |
| User edits → calls `reset()` | `true` | `true` |

The biggest visible difference: a pre-filled form is now in its initial state. The cascade from that is handled deliberately in Decisions 4 and 5.

**Trade-off / pushback room.** Consumers using "any field has a value" as a proxy for "user has interacted" will see different behavior on pre-filled forms. The new signal is closer to what they almost certainly actually wanted.

> **Reviewer ask:** Sign off that "user has edited" (HTML `dirtyValueFlag`) is the right definition for `isInitialState`?

---

### Decision 3 — `form.validity` stays gated on `isInitialState`

**Decision.** The public `validity` getter continues to return `null` while the form is in its initial state. Only after the form transitions out of initial state is validity actually computed.

**Why.** This is a UX contract consumers depend on. Returning `null` means "form has not yet been engaged; don't show errors yet." Removing the gate would make pre-filled forms with empty required fields immediately report as `'invalid'`, which would cascade into error styling and red text before the user has had a chance to do anything.

**What this means.** A pre-filled valid form will report `form.validity === null` at first render even though it's structurally valid. This is intentional. Submission (Decision 4) bypasses this gate via a separate mechanism so that the gate doesn't lock out programmatic submission.

**Trade-off / pushback room.** Consumers that check `form.validity === 'valid'` to enable downstream UI will see `null` at first render even when nothing is wrong. They'd need to either also treat `null` as "no errors" or use a different signal.

> **Reviewer ask:** Is keeping the `validity === null` gate the right consumer-facing contract, or should validity reflect raw constraint state from the start?

---

### Decision 4 — Submit button enablement and `submit()` use raw validity, not gated

**Decision.** The submit button is enabled when no enabled field has a constraint violation, regardless of `isInitialState`. `form.submit()` and Enter-key submission gate on the same raw check. Implemented as a private `_isFormValid()` helper used by all three.

**Why.** The driving scenario: a logged-in user lands on a form pre-populated with their loyalty profile data. They want to click Submit and continue. If submit enablement gated on `isInitialState` (which is now `true` for pre-filled forms per Decision 2), they would be locked out and have to fake an edit. Native HTML doesn't have this concept — Submit is enabled as long as the form is structurally valid.

**What this enables.**
- Pre-filled valid form: Submit button enabled, `form.submit()` fires, Enter submits.
- Empty form with required fields: Submit disabled (constraint violated).
- Empty form with only optional fields: Submit enabled (no constraints to violate).
- Disabled + required field as the only blocker: Submit enabled (disabled fields are skipped per Decision 1).

**Trade-off / pushback room.** Empty optional forms now have submit enabled at first render, which previously was disabled. If the team wanted "user must demonstrate intent before submitting," that signal is gone — but it was based on accidental coupling, not a deliberate policy.

> **Reviewer ask:** Sign off that submit availability should depend on validity alone, not on user interaction?

---

### Decision 5 — Reset button: "form is dirty" OR "any non-disabled field has values/initials"

**Decision.** The Reset button is enabled when either of these is true:
1. The form has diverged from its initial state (`isInitialState === false`), OR
2. Any non-disabled field has a current value or a captured initial value.

Implemented as a private `_hasResetableState()` helper.

**Why.** Reset is meaningful whenever clicking it would change the form's state — either by clearing a current value or by restoring a captured default. Two conditions cover the full space:

- The first condition (form is dirty) catches the edge case where a user has edited a field that subsequently became disabled. Without this branch, the only dirty field is filtered out by the "non-disabled" check in condition 2, and the user is left with a form that knows it's dirty but no UI path to return it to defaults.
- The second condition covers everything else — pre-filled forms have defaults to restore, a user who cleared a pre-filled field back to empty has a default to restore, and an empty untouched form has neither and correctly disables Reset.

**What this enables.**

| Scenario | Reset state | Reason |
|---|---|---|
| Empty form, no values anywhere | Disabled | Neither condition true. |
| Pre-filled form, no edits | Enabled | Initial values present (condition 2). |
| User typed into an empty field | Enabled | Form is dirty (condition 1). |
| User cleared a pre-filled field back to empty | Enabled | Form is dirty + initial value still captured. |
| User edits one field, JS then disables it (the edge case) | **Enabled** | Form is dirty (condition 1) — even though the only field with a value is disabled. |
| Form where only disabled fields have values | Disabled | No non-disabled fields with values; form not dirty (current === initial throughout). |

**Alternatives considered.**
- *Always enable Reset* (mirror native HTML, which never disables it). Simpler — no edge cases to chase. Rejected for keeping the "nothing to do on a truly empty form" UX polish.
- *Tie to `isInitialState` alone*. Rejected — would disable Reset on pre-filled forms.
- *"Any non-disabled field has values/initials" alone* (an earlier version of this decision). Rejected — produces the edge case where a user-edited-then-locked form has Reset grayed out despite the form being dirty.

**Trade-off / pushback room.** The two-condition rule is slightly more complex than either of the rejected alternatives. A reviewer who values rule simplicity above edge-case coverage might prefer the "always enabled" native option. Pre-PR behavior was approximately this same model (it also kept Reset enabled when any field had a value), so this decision is in spirit returning to that with the dirt-tracking refined.

> **Reviewer ask:** Sign off that Reset should follow "form is dirty OR there's something to clear/restore" — rather than collapsing to always-enabled or to `isInitialState` alone?

---

### Decision 6 — Focus/blur on an optional field no longer taints the form (out of scope, kept)

**Decision.** This is a side effect of the new `_setInitialState` predicate, not a goal. Kept rather than reverted, because it aligns with native HTML behavior.

**The code path, confirmed from `base-input.js` and `validation.js`:**
1. `handleFocusin()` sets `this.touched = true`.
2. `handleBlur()` calls `this.validation.validate(this)`.
3. `auro-input.value` is **not** mutated by focus/blur — the host property stays `undefined` for a never-typed field. (The inner native `<input>.value` is `''` by default, but the host property is untouched.)
4. Because no value changed, no `input` event fires; `formState[name].value` is not updated.
5. `validate()` sets `this.validity` to `'valid'` (optional) or `'valueMissing'` (required) and dispatches `auroFormElement-validated`. The form writes that into `formState[name].validity`.

**Resulting form state after pure focus → blur:**

| Field type | `formState[x]` after focus+blur | `dev` `isInitialState` | This PR `isInitialState` | Same? |
|---|---|---|---|---|
| Optional | `{ value: null, validity: 'valid' }` | `false` (validity ≠ null tainted) | `true` (value unchanged, `'valid'` not an error) | **Different** |
| Required | `{ value: null, validity: 'valueMissing' }` | `false` | `false` (error validity taints) | Same |

The shift is narrow — **optional fields only**. Required-field focus/blur still taints because `'valueMissing'` is an error validity, which the new predicate counts.

**Downstream effects (optional-field focus+blur only):**
- `form.isInitialState`: was `false`, now `true`.
- `form.validity`: was `'valid'`, now `null` (gated per Decision 3).
- Reset button: was enabled (form tainted), now disabled (no values, no initials).
- Submit button, `submit()`, Enter: behavior unchanged from a user perspective (all enabled / firing).
- Individual field error UI on the input itself: unchanged.

**Why kept.** Native HTML doesn't treat traversing focus through a control as a meaningful event for form-level state. The previous behavior was an incidental consequence of the old predicate, not a deliberate design. Reverting it would require keeping the OR-on-non-null-validity term in `_setInitialState`, which would also re-introduce the "default-valued field is always tainted" problem that Decisions 2 and 4 are designed to fix.

**What could break for consumers.**
1. Code using `form.isInitialState === false` as a "user has engaged" signal. Migration: per-field `touched` state on each `auro-input`, or listen for `auroFormElement-validated` events from form children.
2. Code reading `form.validity === 'valid'` as "ready to act." Migration: also handle `null`, or use the form's `submit` event.

> **Reviewer ask:** Acknowledge that this side effect is being shipped under the v6 breaking window with platform-alignment rationale? (No active action needed — just a sign-off that this is documented and accepted.)

---

### Other items worth mentioning (no decision needed)

- **`disconnectedCallback`.** The form previously had no `disconnectedCallback`, which meant existing `mutationObservers` and the new `_attributeObserver` would leak. The PR adds cleanup. Pure bug fix.
- **`MutationObserver` for `disabled` and `name`.** Picks up runtime attribute toggles so the form's view of who-is-disabled stays current.
- **`FormState` JSDoc.** Removed an unused `element` property from the typedef. Docs only.
- **Test `should disable buttons when form is in initial state`.** Fixture changed from optional-empty to required-empty. Empty optional forms now correctly enable Submit (Decision 4); the test's intent — "both buttons disabled when there's nothing to do" — is preserved with the required-field fixture.

---

### Summary of the consumer-facing API contract after this PR

| Surface | Contract |
|---|---|
| `form.value` | Object keyed by field `name`, excludes disabled fields. |
| `form.validity` | `'valid'`, `'invalid'`, or `null`. Stays `null` while the form is in its initial state. |
| `form.isInitialState` | `true` until the user edits a field (current value diverges from captured initial) or a field enters an error validity state. Disabling a previously-edited field does **not** clear this. |
| `form.submit()` | Validates enabled fields, dispatches `submit` event when raw constraint validation passes. Works on pre-filled valid forms with no prior interaction. |
| Submit button enablement | Enabled when raw constraint validation passes. Disabled when any enabled field has an unmet constraint. |
| Reset button enablement | Enabled when the form is non-initial OR when any non-disabled field has a current value or captured initial value. |
| Enter key on a form element | Submits via `form.submit()` unless the source is disabled or a textarea. |

---

### Sign-off

Once you've reviewed:
- ✅ / ❌ Decision 1 (disabled-control alignment)
- ✅ / ❌ Decision 2 (dirty-value-flag `isInitialState`)
- ✅ / ❌ Decision 3 (gated `validity` getter)
- ✅ / ❌ Decision 4 (raw validity for submission paths)
- ✅ / ❌ Decision 5 (resetable-state for Reset button)
- ✅ / ❌ Decision 6 (focus/blur side effect kept)

Feedback or pushback on any decision is welcome — every section is meant to be independently challengeable.


components/form/src/auro-form.js
```diff
@@ -169,6 +176,28 @@ export class AuroForm extends LitElement {
...
+  /**
+   * Whether a given element is currently disabled. Disabled controls are excluded
+   * from submission, validity, and initial-state checks (per the HTML spec).
+   * @param {HTMLElement | undefined | null} element - The element to check.
+   * @returns {boolean}
+   * @private
+   */
+  _isDisabled(element) {
+    return Boolean(element?.hasAttribute('disabled'));
+  }
```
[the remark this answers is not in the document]
jordanjones243: Not fixed — documented in a code comment on `_isDisabled`. Reason: every Auro component in `formElementTags` declares disabled with reflect: true, so attribute and property stay in sync — checking the attribute is sufficient. The fieldset-inheritance suggestion is a real native-HTML behavior, but no precedent exists in this codebase for walking ancestors and the ticket didn't ask for it. The comment tells the next reader exactly when to expand the helper (any future form-element type that ships without reflect).

jordanjones243:
PR is blocked until it can be included in the Formkit v6 release. Tagging @rmenner and @jason-capsule42 for awareness of this ticket needing to be bundled into that release.

components/form/src/auro-form.js
```diff
@@ -242,6 +295,10 @@ export class AuroForm extends LitElement {
       // go through validity states and return the first invalid state (if any)
       const invalidKey = Object.keys(this.formState).
         find((key) => {
+          if (this._isNameDisabled(key)) {
+            return false;
+          }
```
jason-capsule42: Please add a comment here that explains disabled forms are not included in form data nor block submission per section `4.10.19.2 Enabling and disabling form controls` of https://www.w3.org/TR/2011/WD-html5-20110113/association-of-controls-and-forms.html

components/form/src/auro-form.js
```diff
@@ -242,6 +295,10 @@ export class AuroForm extends LitElement {
       // go through validity states and return the first invalid state (if any)
       const invalidKey = Object.keys(this.formState).
         find((key) => {
+          if (this._isNameDisabled(key)) {
```
jason-capsule42: Would line 260 above have already filtered out the disabled form field making this redundant?

components/form/src/auro-form.js
```diff
@@ -268,7 +325,13 @@ export class AuroForm extends LitElement {
    * @private
    */
   _setInitialState() {
-    const anyTainted = Object.keys(this.formState).some((key) => this.formState[key].validity !== null || this.formState[key].value !== null);
+    const anyTainted = Object.keys(this.formState).some((key) => {
+      if (this._isNameDisabled(key)) {
+        return false;
+      }
```
jason-capsule42: What is the purpose of this check? If the the form field was disabled at first render it would never be tainted. If it was disabled later, after a user set a value, shouldn't that still report the the the form was not in it's initial state?

components/form/src/auro-form.js
```diff
@@ -435,10 +497,13 @@ export class AuroForm extends LitElement {
...
-    // Force validation on ALL elements
-    this._elements.forEach((element) => {
-      element.validate(true);
-    });
+    // Force validation on all enabled elements. Disabled fields are skipped
+    // because disabled controls are not validated nor submitted per the HTML spec.
+    this._elements.
+      filter((element) => !this._isDisabled(element)).
+      forEach((element) => {
+        element.validate(true);
```
jason-capsule42, line 637: What case does revalidation on submit cover? Every field already validates any time the value changes. Is there a case where a form field value can become invalid but hasn't run validity yet? Would doing this again here just make the form less performant without adding any benefit?

components/form/src/auro-form.js
```diff
@@ -526,6 +591,11 @@ export class AuroForm extends LitElement {
    */
   handleKeyDown(event) {
     if (event.key === 'Enter' && this.isFormElement(event.target)) {
+      // Disabled controls do not submit a form natively.
+      if (this._isDisabled(event.target)) {
+        return;
+      }
```
jason-capsule42, line 738: How would an enter key event originate from a disabled form field?

components/form/src/auro-form.js
```diff
@@ -557,6 +627,81 @@ export class AuroForm extends LitElement {
     });
   }
 
+  /**
+   * Handle batched MutationObserver records for `disabled` and `name`
```
jason-capsule42, line 772: Should we also be doing `validity` state changes in addition to `disabled` and `name`?

jason-capsule42 (commented):
I left various questions in the review.

components/form/demo/customize.js
```diff
@@ -1,9 +1,26 @@
...
+export async function initExamples(initCount) {
+  initCount = initCount || 0;
+
+  try {
+    await disabledExample();
+  } catch (err) {
+    if (initCount <= 20) {
+      setTimeout(() => {
+        initExamples(initCount + 1);
+      }, 100);
```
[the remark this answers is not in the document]
jordanjones243: This is a standard followed in every other formkit component, whether right or wrong.

components/form/src/auro-form.js
jordanjones243: Line 260 is inside the `value` getter — it filters disabled fields out of the submitted payload. Line 298 is inside `_calculateValidity`, which iterates `formState` directly (not via `value`) and needs its own guard to skip disabled fields when deciding whether the form is valid. Same predicate (`_isNameDisabled`), different code paths — not redundant.

components/form/src/auro-form.js
jordanjones243: The case it covers is a required field that's never been touched. `validate(true)` (with the force flag) runs even when `validity === null`, transitioning it to `'invalid'` for the `valueMissing` case. Without this pre-submit pass, an untouched required field stays at `validity === null`, the form's overall `validity` resolves to `'valid'`, and submit goes through with missing data. The cost is one validate call per enabled field at submit time — negligible compared to the user pressing Submit, and it's the only place that catches "user never touched a required field."

components/form/src/auro-form.js
jordanjones243: Fair question — for the common case it can't, because auro-input reflects `disabled` to its inner native `<input>` (which then can't receive focus, so no keydown bubbles up). I left the guard in for two reasons:

1. Not every component in `formElementTags` necessarily delegates `disabled` to a focusable native element the same way auro-input does — radios, checkboxes, counter, datepicker, combobox each have their own shadow DOM. A defensive check here keeps the form's behavior uniform regardless of how a given component handles disabled.
2. Programmatically dispatched `KeyboardEvent`s bypass native disabled gating entirely, so the check protects against that path too.

The cost is one boolean check and it matches the consistent disabled-skip pattern used at every other entry point in this file. Happy to remove it if you'd rather not carry the defensive code, but I'd lean toward keeping.

components/form/src/auro-form.js
jordanjones243: `validity` is reflected as an attribute on the form elements (it's declared `reflect: true` in `BaseInput`), so technically the observer could watch it — but it'd be redundant. Every validity change already fires `auroFormElement-validated`, which `sharedValidationListener` listens for and uses to update `formState[name].validity`. Adding `validity` to the attribute observer would double-fire (event + mutation) on every validation cycle.

The reason `disabled` and `name` need the observer is that neither has an equivalent dedicated event — the observer is the only signal we have. `validity` already has its own first-class channel.

components/form/test/auro-form.test.js
```diff
@@ -649,6 +649,218 @@ function runFullTest(mobileView) {
...
+          <auro-input name="toggleField" value="hello" disabled></auro-input>
+        </auro-form>
+      `);
+      await elementUpdated(el);
+
+      expect(el.value).to.not.have.property('toggleField');
+
+      const inputEl = el.querySelector('auro-input[name="toggleField"]');
+      inputEl.removeAttribute('disabled');
+      await elementUpdated(el);
```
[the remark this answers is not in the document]
jordanjones243: The assertion on `el.value` doesn't depend on the observer firing — `value` is a getter that calls `_isNameDisabled(key)` on every read, and `_isNameDisabled` checks `hasAttribute('disabled')` live against the current DOM. So as soon as `removeAttribute('disabled')` returns, the next read of `el.value` reflects the change regardless of whether the MutationObserver has run.

The observer's job is to trigger a re-render and refresh button-disabled state — not to gate the getter. The test's `await elementUpdated(el)` is for ergonomics (letting any pending render finish) rather than correctness. So it shouldn't be flaky, but if you'd like I can add a comment explaining why the bare getter is sufficient here.

components/form/src/auro-form.js
jordanjones243: Good call — adopted. Removed the `_isNameDisabled` short-circuit in `_setInitialState` and switched to a per-field initial-value comparison (HTML's `dirtyValueFlag` model). The exact case you flagged — user edits a field, then JS disables it — now correctly stays tainted. Details and the full set of cascading decisions are in the **Behavioral decisions (round 2 review)** section of the updated PR description, particularly Decision 2.

components/select/demo/customize.html
```diff
@@ -41,12 +41,12 @@
...
     <script src="https://cdn.jsdelivr.net/npm/@aurodesignsystem/auro-button@latest/+esm" type="module"></script>
-    <script type="module">
+    <script type="module" data-demo-script="true">
       import { renderPage } from './demo-support.min.js';
       await renderPage('./customize.md');
       import { initExamples } from "./customize.min.js";
       initExamples();
     </script>
-    <script src="./customize.min.js" data-demo-script="true" type="module"></script>
+    <script src="./customize.min.js" type="module"></script>
```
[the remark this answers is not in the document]
jordanjones243: Adopted — removed the redundant `<script src="./customize.min.js">` line. The inline module already imports `initExamples` from the same file, so the second tag was dead weight. Now matches the form demo pattern (`components/form/demo/customize.html`).

components/form/src/auro-form.js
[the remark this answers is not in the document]
jordanjones243: Added — see the new `does not submit when Enter is pressed on a disabled form element` test in `auro-form.test.js`, placed alongside the existing Enter-key suite. It dispatches a programmatic `KeyboardEvent` on a disabled `auro-input` and asserts the form's `submit` event does not fire.

components/form/src/auro-form.js
```diff
@@ -352,9 +495,13 @@ export class AuroForm extends LitElement {
       value: element.value || element.getAttribute('value'),
       validity: element.validity || null,
       required: element.hasAttribute('required'),
-      // element
     };
 
+    // Capture the initial (default) value once per name. The `??=` guard
+    // preserves the original capture across rename/slot/reset cycles so
+    // `_setInitialState` can detect user edits as `current !== initial`.
+    this._initialValues[targetName] ??= this.formState[targetName].value;
```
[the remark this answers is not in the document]
jordanjones243: Adopted — fixed in the rename branch of `_handleAttributeMutations`. The MutationObserver now records `attributeOldValue: true`, and before `initializeState()` re-runs we migrate each renamed field's captured initial value from `_initialValues[oldName]` to `_initialValues[newName]` and delete the stale old-name entry. So the new-name lookup in `_addElementToState` hits the correct preserved initial, `current !== initial` still holds for an edited-then-renamed field, and the form stays non-initial. New regression test in `auro-form.test.js` covers exactly this path.

components/form/src/auro-form.js
```diff
@@ -557,6 +714,82 @@ export class AuroForm extends LitElement {
...
+      // initializeState() rebuilds formState from scratch (re-keying any
+      // renamed element) and also dispatches `change` + refreshes button state.
+      // We also re-run _attachEventListeners() because elements that previously
+      // had no `name` were skipped by queryAuroElements() (which selects
+      // `[name]`) and therefore never received input/validation/keydown
+      // listeners. Re-attaching is safe — the listener-removal step inside
+      // _attachEventListeners() prevents duplicates on already-wired elements.
+      this.initializeState();
+      this._attachEventListeners();
+      return;
```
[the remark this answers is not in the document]
jordanjones243: Adopted — exactly as you suggested. Added `attributeOldValue: true` to the observer options, and the rename branch of `_handleAttributeMutations` now migrates `_initialValues[oldName] -> _initialValues[newName]` (with the old key deleted) before `initializeState()` re-keys formState. This also stops the unbounded growth of `_initialValues` across repeated renames. New regression test in `auro-form.test.js` verifies an edited-then-renamed field keeps the form non-initial and Reset enabled.

jordanjones243:
Please note that the test failing in `auro-menu` are pre-exiting issues from the `breaking/formkit-v6` branch

components/form/src/auro-form.js
```diff
@@ -231,23 +299,71 @@ export class AuroForm extends LitElement {
...
+    }
+
+    return Object.keys(this.formState).some((key) => {
+      if (this._isNameDisabled(key)) {
+        return false;
+      }
+      const current = this.formState[key].value;
+      const initial = this._initialValues[key] ?? null;
+      return current !== null || initial !== null;
+    });
```
jason-capsule42, line 360: Does this mean Reset doesn't affect disabled fields?

components/form/src/auro-form.js
```diff
@@ -294,18 +429,26 @@ export class AuroForm extends LitElement {
...
   setDisabledStateOnButtons() {
     this._resetElements.forEach((element) => {
-      if (this.isInitialState) {
-        element.setAttribute("disabled", "");
-      } else {
+      // Reset is meaningful whenever any non-disabled field has a current
+      // value OR a captured default value — i.e., whenever the click would
+      // either clear something or restore a default.
+      if (this._hasResetableState()) {
         element.removeAttribute("disabled");
```
jason-capsule42: Should we just modify the disabled prop here?

components/form/src/auro-form.js
```diff
@@ -264,11 +380,30 @@ export class AuroForm extends LitElement {
...
    */
   _setInitialState() {
-    const anyTainted = Object.keys(this.formState).some((key) => this.formState[key].validity !== null || this.formState[key].value !== null);
+    const anyTainted = Object.keys(this.formState).some((key) => {
+      const initialValue = this._initialValues[key] ?? null;
+      const currentValue = this.formState[key].value;
+      const fieldValidity = this.formState[key].validity;
+      // eslint-disable-next-line no-extra-parens
+      return currentValue !== initialValue || (fieldValidity !== null && fieldValidity !== 'valid');
+    });
```
jason-capsule42, line 449: Need to understand this logic better.

components/form/src/auro-form.js
jason-capsule42: AI answer:

Yes, setting the property is more idiomatic and robust for web components. [element.disabled = true](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) works directly with Lit's reactive property system, while [setAttribute](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)/[removeAttribute](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) relies on the component's attribute-changed callback to sync the property. For native <button> elements it's equivalent, but for auro-button (a Lit element), the property setter is the more direct path.

components/form/src/auro-form.js
jason-capsule42: the JSDoc comment above this method says "we deliberately do NOT treat [validity !== null](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) as a taint signal," but the code does taint on [validity !== null && validity !== 'valid'](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html). The comment is slightly misleading — it's saying that [validity === 'valid'](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) alone doesn't taint (since Auro elements auto-validate on mount), but an invalid validity result still does.

jason-capsule42 (dismissed):

components/form/src/auro-form.js
jordanjones243: You're right, the comment was confusing — I rewrote it. Short version: a field counts as tainted if its value changed or it's currently failing validation.

The value check covers the obvious case (user typed something). The validity check is a backup for cases where the user interacted but the value didn't change — like focusing a required field and blurring without typing, which leaves it 'invalid' even though the value is still empty.

We skip null (field hasn't been validated yet) and 'valid' (the default state after Auro auto-validates on mount) because neither one proves the user touched anything.

components/form/src/auro-form.js
```diff
@@ -263,19 +379,33 @@ export class AuroForm extends LitElement {
...
+   *
+   * Validity acts as a backup signal: it catches users who interact with a
+   * field without changing its value (e.g., focusing and blurring a required
+   * field). We skip `null` (not yet validated) and `'valid'` (the default
+   * after Auro's auto-validation on mount) because neither proves the user
+   * touched anything.
    * @returns {void}
    * @private
    */
   _setInitialState() {
```
chrisfalaska, line 438: `_setInitialState` flickers reset-button state during `reset()`

`_setInitialState` unconditionally sets `resetElement.disabled = false` (line ~409), then `setDisabledStateOnButtons` re-decides. In the `reset()` path the second call is deferred behind an extra `updateComplete.then`, allowing a paint where `reset` is visibly enabled when it shouldn't be.

Fix: delete the unconditional enable in `_setInitialState`; let `setDisabledStateOnButtons` own button state as single source of truth, as it already runs right after in `updated()` and `reset()`.

```
  _setInitialState() {
    const anyTainted = Object.keys(this.formState).some((key) => {
      const initialValue = this._initialValues[key] ?? null;
      const currentValue = this.formState[key].value;
      const fieldValidity = this.formState[key].validity;
      // eslint-disable-next-line no-extra-parens
      return currentValue !== initialValue || (fieldValidity !== null && fieldValidity !== 'valid');
    });

    this._isInitialState = !anyTainted;
    // Button state is owned by setDisabledStateOnButtons (called from updated()
    // and reset()). Touching resetElement.disabled here causes a visible flicker
    // in reset(), where the final setDisabledStateOnButtons is deferred behind
    // an extra updateComplete.
  }
```

components/form/src/auro-form.js
chrisfalaska, line 551: `initializeState` rebuilds formState but never prunes `_initialValues` of keys no longer present → unbounded growth in apps with dynamic fields.

After rebuilding `formState`, drop captured initials for fields that no longer exist. This bounds growth in apps with dynamic fields without breaking rename/slot/reset preservation (which all run before this prune).

```
  initializeState() {
    this.formState = {};
    this._submitElements = [];
    this._resetElements = [];
    this._elements = [];

    this.queryAuroElements().forEach((element) => {
      // ...unchanged...
    });

    // Drop captured initial values for fields that no longer exist in the form.
    // Rename migration in _handleAttributeMutations has already re-keyed surviving
    // fields, so anything left here is a field that was removed from the DOM.
    for (const key of Object.keys(this._initialValues)) {
      if (!(key in this.formState)) {
        delete this._initialValues[key];
      }
    }

    this.dispatchEvent(new Event('change', { /* ... */ }));
    this.setDisabledStateOnButtons();
  }
```

components/form/src/auro-form.js
```diff
@@ -557,6 +705,97 @@ export class AuroForm extends LitElement {
...
+        'disabled',
+        'name'
+      ]
+    });
+  }
+
+  /**
+   * @returns {void}
+   */
+  disconnectedCallback() {
```
chrisfalaska, line 872: Don't wipe `_initialValues` on disconnect

Clearing the captured baselines on disconnect causes the form to silently re-baseline against the user's edited values after a slot-move/reattach.

Keep the observer cleanup; drop the `_initialValues` reset.

```
  disconnectedCallback() {
    // Disconnect everything we own to avoid leaking observers (and the strong
    // refs they hold to DOM nodes) past the form's lifetime.
    this._attributeObserver?.disconnect();
    this.mutationObservers.forEach((observer) => observer.disconnect());
    this.mutationObservers = [];
    // Intentionally do NOT clear _initialValues here. Slot-moves trigger a
    // disconnect/reconnect cycle; clearing would cause the next initializeState
    // to capture the user's edited values as the new "initial" and silently
    // flip the form back to its initial state.

    super.disconnectedCallback();
  }
```

components/form/src/auro-form.js
```diff
@@ -557,6 +705,97 @@ export class AuroForm extends LitElement {
...
+
+    const renameMutations = relevant.filter((mutation) => mutation.attributeName === 'name');
+    if (renameMutations.length > 0) {
+      // Migrate each renamed field's captured initial value from the old key
+      // to the new key before `initializeState()` re-runs `_addElementToState`.
+      // Without this, the new-name lookup in `_initialValues` would miss, the
+      // field's current (possibly user-edited) value would be captured as the
+      // new initial, and the form would incorrectly flip back to its initial
+      // state. The old key would also leak in `_initialValues` indefinitely.
+      renameMutations.forEach((mutation) => {
```
chrisfalaska, line 799: When `getAttribute('name')` returns `null` (attribute removed), drop the orphan key instead of silently skipping.

```
  renameMutations.forEach((mutation) => {
    const oldName = mutation.oldValue;
    const newName = mutation.target.getAttribute('name');
    if (!oldName || oldName === newName) {
      return;
    }
    if (newName === null) {
      // `name` attribute removed — field will fall out of formState on re-init.
      // Drop its captured initial so it doesn't leak in _initialValues.
      delete this._initialValues[oldName];
      return;
    }
    if (oldName in this._initialValues) {
      this._initialValues[newName] = this._initialValues[oldName];
      delete this._initialValues[oldName];
    }
  });
```

components/form/src/auro-form.js
chrisfalaska, line 521: Store `disabled` directly on each `formState` entry.  Update it from the observer. Eliminates the `_elements.find()` and the `O(n²)` cost on every read.

```
  _addElementToState(element) {
    const targetName = element.getAttribute('name');
    if (this.formState[targetName]) {
      return;
    }

    this.formState[targetName] = {
      value: element.value || element.getAttribute('value'),
      validity: element.validity || null,
      required: element.hasAttribute('required'),
      disabled: element.hasAttribute('disabled'),
    };

    if (!(targetName in this._initialValues)) {
      this._initialValues[targetName] = this.formState[targetName].value;
    }

    this._elements.push(element);
  }

  _isNameDisabled(name) {
    return Boolean(this.formState[name]?.disabled);
  }
```

Then in `_handleAttributeMutations`, refresh the cached flag for non-rename mutations before the re-render:

```
  // Non-rename branch (`disabled` toggle on a tracked field)
  relevant
    .filter((mutation) => mutation.attributeName === 'disabled')
    .forEach((mutation) => {
      const name = mutation.target.getAttribute('name');
      if (name && this.formState[name]) {
        this.formState[name].disabled = mutation.target.hasAttribute('disabled');
      }
    });

  this.requestUpdate('formState');
  this.setDisabledStateOnButtons();
```

  Note: this removes the load-bearing reliance on `_elements` from the hot path, so the `_isNameDisabled` doc comment about "future form-element type ships without attribute
  reflection" can stay accurate — the cache is fed by the observer, which reads the attribute the same way.

chrisfalaska (changes requested):

chrisfalaska:
Copy/paste-ready Markdown:

````md
## Suggested Test Additions

Drop these into `components/form/test/auro-form.test.js`:

```js
it('does not flicker reset button enablement during reset()', async () => {
  // Render form with a pre-filled field + reset button, edit, then reset.
  // Assert resetButton.disabled is never true→false→true within one frame
  // by snapshotting after reset() but before awaiting updateComplete.
});

it('prunes _initialValues when a tracked field is removed from the DOM', async () => {
  // Render form with [name="a"], assert _initialValues.a exists, remove the
  // node, trigger slotchange / initializeState, assert _initialValues.a is gone.
});

it('preserves _initialValues across a disconnect/reconnect cycle', async () => {
  // Render, capture initial, edit, detach the host from the DOM, re-attach.
  // Assert isInitialState === false still holds.
});

it('drops _initialValues for a field whose name attribute is removed', async () => {
  // Render with [name="x"], edit, removeAttribute('name'), await updateComplete.
  // Assert _initialValues.x is gone and form.value has no 'x' key.
});

it('rename collision (two fields → same name) does not leak _initialValues', async () => {
  // Render with [name="a"] and [name="b"], rename b → a, await updateComplete.
  // Assert Object.keys(_initialValues) has exactly the surviving keys.
});

it('handles all-disabled form: empty value, valid validity, submit fires with empty payload', async () => {
  // Pre-existing _isFormValid() returns true for no enabled fields.
});
```

---

## Suggested Documentation Additions

In `components/form/docs/pages/customize.md`, add a short subsection under the v6 behavior changes covering:

- **Runtime renames:** Changing `name` on a tracked field re-keys form state and preserves the captured initial value.
- **Late name attachment:** Elements that gain a `name` attribute after first render are picked up automatically.
- **Reset gating:** Reset is enabled whenever the form is dirty, or any non-disabled field carries a current or default value.
- **Enter-to-submit on disabled fields:** Suppressed (matches native behavior).
````

File: components/form/src/auro-form.js (hunk 16 above)
Comment (jordanjones243) [reply components/form/src/auro-form.js:438]: Addressed.

File: components/form/src/auro-form.js
Comment (jordanjones243) [reply components/form/src/auro-form.js:551]: Addressed.

File: components/form/src/auro-form.js (hunk 17 above)
Comment (jordanjones243) [reply components/form/src/auro-form.js:872]: Addressed.

File: components/form/src/auro-form.js (hunk 18 above)
Comment (jordanjones243) [reply components/form/src/auro-form.js:799]: Addressed.

File: components/form/src/auro-form.js
Comment (jordanjones243) [reply components/form/src/auro-form.js:521]: Addressed.

File: components/form/src/auro-form.js [hunk 19]
```diff
@@ -17,7 +17,6 @@ import AuroLibraryRuntimeUtils from '@aurodesignsystem/auro-library/scripts/util
  * @property {string | number | boolean | string[] | null} value - The value of the form element.
  * @property {ValidityState} validity - The validity state of the form element, stored when fired from the form element.
  * @property {boolean} required - Whether the form element is required or not.
- * @property {HTMLElement} element - Whether the form element is required or not.
  */
```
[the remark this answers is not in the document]
jordanjones243: Addressed.

components/form/src/auro-form.js
```diff
@@ -169,6 +193,45 @@ export class AuroForm extends LitElement {
...
+   * element's `disabled` attribute toggles. The cache is fed by the same
+   * `hasAttribute('disabled')` read as `_isDisabled`, so the "future form-element
+   * type without attribute reflection" caveat documented there applies here too.
+   * @param {string} name - The `name` attribute used to register the element.
+   * @returns {boolean}
+   * @private
+   */
+  _isNameDisabled(name) {
+    return Boolean(this.formState[name]?.disabled);
+  }
```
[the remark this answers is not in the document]
jordanjones243: The cache is intentional and `submit()` is already consistent within a tick — the validate filter reads live via `_isDisabled(element)`, then `await this.updateComplete` resolves after pending microtasks (which is when MutationObservers fire), so by the time `_isFormValid()` and `this.value` read the cache it's already refreshed. No mixed-source-of-truth bug at submit time.

The "read live, fall back to cache" approach defeats the perf goal of the cache: looking up the live attribute by `name` requires `_elements.find(el => el.getAttribute('name') === name)` per call, which puts us back at O(n²) for `value`/`validity`/`isInitialState` reads — exactly the regression the cache was designed to eliminate. A nameToElement Map would make live reads O(1) but is a larger refactor than what was scoped here.

The remaining stale window is the synchronous read-after-toggle pattern (`field.toggleAttribute('disabled'); form.value`), which is bounded to one microtask. Tests that hit this can `await elementUpdated(form)` — the standard pattern in this codebase. Happy to file a follow-up to convert to a Map-backed live read if anyone reports the sync window biting them in production, but I'd rather not eat the perf cost preemptively.

components/form/src/auro-form.js
```diff
@@ -231,23 +299,113 @@ export class AuroForm extends LitElement {
...
+      // We treat a field as invalid in two cases:
+      //   1. validity is known-bad (non-null, non-'valid')
+      //   2. it's `required` and structurally empty — `valueMissing` is
+      //      certain even without a validation pass.
+      // `submit()` calls `validate(true)` on every enabled field before
+      // reading `_isFormValid()`, so any not-yet-validated field that turns
+      // out to fail another constraint still blocks dispatch.
+      const knownInvalid = formKey.validity !== null && formKey.validity !== 'valid';
+      const requiredAndEmpty = formKey.required && this._normalizeEmpty(formKey.value) === null;
+      return knownInvalid || requiredAndEmpty;
```
chrisfalaska, line 332: The previous logic treated any field with a value and `validity !== 'valid'` as invalid — that included `validity === null`.

The new code only blocks on `knownInvalid` (non-null, non-`'valid'`) or `requiredAndEmpty`. An optional field whose value fails a format constraint but whose validity is still `null` (for example, framework gating suppressed the dispatch — see line 723 where we `?? null` in exactly that case) now passes `_isFormValid()`.

The inline comment argues that `submit()` calls `validate(true)` before reading this, which surfaces failing constraints. But if `validate(true)` hits the same gating that produced the undefined validity in the first place, an optional invalid field could slip through.

Can we add a regression test for an optional field with a malformed but non-empty value and verify that submit is blocked? If it isn't, we should fall back to the previous behavior for non-required fields that contain a value.

>line 723 `(event.detail.validity ?? null)` -- the existence of this normalization is the evidence that `validity === null` after a validation pass is a real state, which is what makes the concern above more than theoretical.

## Suggested Regression Test

Place this inside `describe('submit', () => { ... })` immediately after the existing `"does not dispatch submit event when form is invalid"` test at `components/form/test/auro-form.test.js:1316`.

```js
// Regression guard: an OPTIONAL field with a malformed value (validity
// like `'patternMismatch'`, not `null`/`'valid'`) must still block submit.
// The risk pinned in _isFormValid (auro-form.js:330-332) is that the
// rewrite only blocks on `knownInvalid || (required && empty)`. If
// `validate(true)` inside submit() leaves validity at `null` for a non-
// required field — and sharedValidationListener's `?? null` at line 723
// proves null-after-validate is a real state — the submit gate misses it.
//
// auro-input flips type="email" with a bad value to validity='patternMismatch'
// (see components/input/test/auro-input.test.js:1448), so the test asserts
// the form correctly reads that and blocks dispatch.
it('does not submit when an optional field has invalid (non-empty) input', async () => {
  const el = await fixture(html`
    <auro-form>
      <auro-input
        name="optionalEmail"
        type="email"
        value="not-an-email">
      </auro-input>
    </auro-form>
  `);

  await elementUpdated(el);

  let submitFired = false;

  el.addEventListener('submit', () => {
    submitFired = true;
  });

  await el.submit();

  // If this fails, _isFormValid is too permissive on optional fields:
  // revert that branch to the prior shape for non-required-with-value:
  //   (validity !== 'valid' && required)
  //     || (validity !== 'valid' && value !== null)
  expect(
    submitFired,
    'submit should be blocked on optional field with invalid value'
  ).to.be.false;

  expect(
    el._isFormValid(),
    '_isFormValid should reject the malformed value'
  ).to.be.false;
});
```

## Likely Outcome

Based on the current codebase, `auro-input`'s `updated()` lifecycle runs validation when a value is set. By the time `submit()` calls `validate(true)`, the input's `.validity` should already be `'patternMismatch'`.

The form should therefore see:

```js
knownInvalid === true
```

and block submission. The test should pass, which confirms the new `_isFormValid()` implementation behaves correctly and locks that behavior in against future regressions.

## If the Test Fails

A failure would likely look like:

```js
submitFired === true
```

In that case, revert the `_isFormValid()` branch at `auro-form.js:332` to:

```js
return (formKey.validity !== 'valid' && formKey.required) ||
       (formKey.validity !== 'valid' && formKey.value !== null);
```

## Optional Belt-and-Suspenders Check

To explicitly cover the "value set as an attribute, never typed into" path, add a sanity check before `submit()`:

```js
const input = el.querySelector('auro-input[name="optionalEmail"]');

expect(input.getAttribute('validity')).to.equal('patternMismatch');
```

If this assertion fails, validity is not being established during registration for attribute-driven values. That would point directly at the edge case under discussion, and the next thing to verify would be whether `validate(true)` inside `submit()` completes synchronously enough to populate validity before `_isFormValid()` runs.

chrisfalaska (changes requested):

components/form/src/auro-form.js
jordanjones243: Added.

chrisfalaska (approved):
Looks great! 🚀

PR #1503, opened 2026-06-12
ci: fix packaging issue AB#1575423

sun-mota:

## Review Checklist:

<details>
<summary>
RC Checklist
</summary>

### Browsers

Android

- [ ] Chrome
- [ ] Firefox

 iOS

- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Desktop

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### Scenarios

- [ ] Next React
- [ ] SvelteKit

</details><br>

-- Auro Design System Team

## Summary by Sourcery

Build:
- Enable Rollup to prefer Node built-in modules in the internal bundling configuration.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -13,6 +13,9 @@ jobs:
     with:
       demo-dir: './auro-formkit-demo'
       demo-build-command: 'npm run local-demo'
+      # test:dist reads components/*/dist/, so auro-actions must run
+      # additional-build-command AFTER `npm run build`. If that order
+      # changes upstream, the check stops catching anything.
       additional-build-command: 'npx playwright install chromium && npm run test:dist'
       cache-dirs: |
         ./components/**/dist
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -16,7 +16,11 @@ jobs:
       # test:dist reads components/*/dist/, so auro-actions must run
       # additional-build-command AFTER `npm run build`. If that order
       # changes upstream, the check stops catching anything.
-      additional-build-command: 'npx playwright install chromium && npm run test:dist'
+      additional-build-command: |
+        npx playwright install chromium
+        echo "::group::Bundled imports verification"
+        npm run test:dist
+        echo "::endgroup::"
       cache-dirs: |
         ./components/**/dist
         ./components/**/demo
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -13,7 +13,7 @@ jobs:
     with:
       demo-dir: './auro-formkit-demo'
       demo-build-command: 'npm run local-demo'
-      additional-build-command: 'npx playwright install chromium'
+      additional-build-command: 'npx playwright install chromium && npm run test:dist'
       cache-dirs: |
         ./components/**/dist
         ./components/**/demo
--- a/packages/config/src/internal.rollup.config.mjs
+++ b/packages/config/src/internal.rollup.config.mjs
@@ -19,7 +19,10 @@ const createConfig = (input, output) => ({
     dir: output,
     entryFileNames: '[name].js'
   },
-  external: EXTERNAL_PACKAGE_NAMES.map((name) => new RegExp(`node_modules/${name}`)),
+  // Anchor the package name to a path boundary so e.g. `lit` doesn't also
+  // externalize a hypothetical `lit-vaadin-helpers`. test/check-bundled-imports
+  // uses an anchored regex on the same allowlist; keep both gates aligned.
+  external: EXTERNAL_PACKAGE_NAMES.map((name) => new RegExp(`node_modules/${name}(?:/|$)`)),
   // Rollup warns and externalizes unresolved imports by default; for published
   // bundles, that means silently shipping broken dist files.
   onwarn(warning, warn) {
--- a/packages/config/src/internal.rollup.config.mjs
+++ b/packages/config/src/internal.rollup.config.mjs
@@ -2,6 +2,16 @@
 
 import { nodeResolve } from '@rollup/plugin-node-resolve';
 
+// Shared by rollup and test/check-bundled-imports.mjs so both gates use the
+// same external-package allowlist.
+export const EXTERNAL_PACKAGE_NAMES = [
+  'lit',
+  'lit-element',
+  'lit-html',
+  '@lit',
+  '@lit-labs',
+];
+
 const createConfig = (input, output) => ({
   input,
   output: {
@@ -9,13 +19,21 @@ const createConfig = (input, output) => ({
     dir: output,
     entryFileNames: '[name].js'
   },
-  external: [
-    /node_modules\/lit/,
-    /node_modules\/lit-element/,
-    /node_modules\/lit-html/,
-    /node_modules\/@lit/,
-    /node_modules\/@lit-labs/,
-  ],
+  external: EXTERNAL_PACKAGE_NAMES.map((name) => new RegExp(`node_modules/${name}`)),
+  // Rollup warns and externalizes unresolved imports by default; for published
+  // bundles, that means silently shipping broken dist files.
+  onwarn(warning, warn) {
+    if (warning.code === 'UNRESOLVED_IMPORT') {
+      throw new Error(
+        `Unresolved import "${warning.source}" in ${warning.importer}. ` +
+        `Possible causes: the dependency is missing from node_modules, ` +
+        `its dist/ entry point has not been built yet ` +
+        `(check turbo.json for a "${warning.source}#build" dependsOn entry), ` +
+        `or the specifier is misspelled.`
+      );
+    }
+    warn(warning);
+  },
   plugins: [
     nodeResolve({
       dedupe: [
@@ -23,7 +41,7 @@ const createConfig = (input, output) => ({
         'lit-element',
         'lit-html'
       ],
-      preferBuiltins: true,
+      preferBuiltins: false,
       moduleDirectories: ['node_modules']
     })
   ]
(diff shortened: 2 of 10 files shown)


test/check-bundled-imports.mjs
```diff
@@ -0,0 +1,65 @@
...
+// packages/config/src/internal.rollup.config.mjs. Any other bare specifier
+// means rollup failed to inline code we meant to ship.
+
+import { readFileSync, existsSync, readdirSync } from 'node:fs';
+import { dirname, join, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { init, parse } from 'es-module-lexer';
+import { EXTERNAL_PACKAGE_NAMES } from '@aurodesignsystem/config/internal.rollup';
+
+const ENTRY_FILES = ['index.js', 'registered.js'];
```
chrisfalaska: Hardcoding ['index.js', 'registered.js'] means a future component that adds a new entry point (or renames one) silently gets skipped. Same class of implicit assumption that caused the original counter bug. Consider deriving entries from each component's package.json:

```
  import { readFileSync, existsSync, readdirSync } from 'node:fs';
  import { dirname, join, resolve } from 'node:path';

  function getEntryFiles(componentDir) {
    const pkgPath = join(componentDir, 'package.json');
    if (!existsSync(pkgPath)) return [];
    const pkg = JSON.parse(readFileSync(pkgPath, 'utf8'));
    const entries = new Set();
    // main field
    if (pkg.main) entries.add(pkg.main.replace(/^\.?\/?dist\//, ''));
    // exports field (handles "." and subpaths)
    if (pkg.exports) {
      for (const value of Object.values(pkg.exports)) {
        const target = typeof value === 'string' ? value : value?.default ?? value?.import;
        if (target?.startsWith('./dist/')) entries.add(target.replace('./dist/', ''));
      }
    }
    return [...entries].filter((f) => f.endsWith('.js'));
  }
```

  Then replace the inner loop:
```
  for (const componentName of listComponentDirs()) {
    const componentDir = join(COMPONENTS_DIR, componentName);
    for (const entry of getEntryFiles(componentDir)) {
      scanned++;
      const failure = checkEntryFile(componentName, entry);
      if (failure) failures.push(failure);
    }
  }
```

.github/workflows/pull-request.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     with:
       demo-dir: './auro-formkit-demo'
       demo-build-command: 'npm run local-demo'
-      additional-build-command: 'npx playwright install chromium'
+      additional-build-command: 'npx playwright install chromium && npm run test:dist'
```
chrisfalaska: AI flagged this in `pull-request` and `release.yml`:

The gate's correctness depends on auro-actions running additional-build-command after npm run build and before actions/cache@v4. That contract isn't pinned or documented here — if the action changes, the gate silently stops gating. Pin the ref and add a comment:

```
  jobs:
    ci:
      # auro-actions@v3 contract: additional-build-command runs AFTER `npm run build`
      # and BEFORE actions/cache@v4. test:dist relies on both — do not unpin without
      # auditing .github/actions/build/action.yml.
      uses: AlaskaAirlines/auro-actions/.github/workflows/PullRequest.yml@v3
      with:
        demo-dir: './auro-formkit-demo'
        demo-build-command: 'npm run local-demo'
        additional-build-command: 'npx playwright install chromium && npm run test:dist'
```
  (Replace @v3 with whatever the current pinned ref actually is.)

.github/workflows/pull-request.yml
```diff
@@ -13,7 +13,7 @@ jobs:
...
       cache-dirs: |
```
chrisfalaska, line 24: AI-flagged:

When this fails in CI, the log just shows a failed additional-build-command step — hard to find. Either run as a separate step in this workflow with a clear name, or wrap the command so the log line is unmistakable:

```
  additional-build-command: |
    npx playwright install chromium
    echo "::group::Bundled imports verification"
    npm run test:dist
    echo "::endgroup::"
```
  The ::group:: syntax gives it a collapsible, named section in the GitHub Actions log.

test/check-bundled-imports.mjs
```diff
@@ -0,0 +1,65 @@
...
+  for (const { label, reason } of failures) console.error(`  ${label}\n    ${reason}\n`);
+  console.error(
+    'These specifiers should have been inlined by rollup. Update\n' +
+    'EXTERNAL_PACKAGE_NAMES in packages/config/src/internal.rollup.config.mjs\n' +
+    'if the change is intentional.\n'
+  );
+  process.exit(1);
+}
+
+console.log(`✓ bundled-imports check passed (${scanned} files scanned)`);
```
chrisfalaska: No test for the verifier itself

new test/check-bundled-imports.test.mjs

A regression in the verifier (broken regex, swallowed error, findUnbundledImports returning empty on parse failure) silently disarms the gate. Fixture-based smoke test:

```
  import { test } from 'node:test';
  import assert from 'node:assert/strict';
  import { init, parse } from 'es-module-lexer';
  import { EXTERNAL_PACKAGE_NAMES } from '@aurodesignsystem/config/internal.rollup';

  const ALLOWED = EXTERNAL_PACKAGE_NAMES.map((name) => new RegExp(`^${name}(?:/.+)?$`));

  function findUnbundledImports(source) {
    const [imports] = parse(source);
    return [...new Set(imports.map((i) => source.slice(i.s, i.e)))]
      .filter((s) => !s.startsWith('.') && !s.startsWith('/'))
      .filter((s) => !ALLOWED.some((re) => re.test(s)));
  }

  await init;

  test('flags bare workspace import', () => {
    const src = `import { AuroDropdown } from '@aurodesignsystem/auro-dropdown';`;
    assert.deepEqual(findUnbundledImports(src), ['@aurodesignsystem/auro-dropdown']);
  });

  test('allows lit and lit subpaths', () => {
    const src = `import { html } from 'lit'; import { ref } from 'lit/directives/ref.js';`;
    assert.deepEqual(findUnbundledImports(src), []);
  });

  test('allows @lit-labs scoped imports', () => {
    const src = `import { Task } from '@lit-labs/task';`;
    assert.deepEqual(findUnbundledImports(src), []);
  });

  test('ignores relative and absolute imports', () => {
    const src = `import './local.js'; import '/abs/path.js';`;
    assert.deepEqual(findUnbundledImports(src), []);
  });
```

Wire into test:dist as a pre-check, or run via node --test.

The cleanest move here is to extract findUnbundledImports + ALLOWED into a small module that both check-bundled-imports.mjs and the test import from.

chrisfalaska (changes requested):

.github/workflows/pull-request.yml
rmenner: Good catch. We can't pin at the moment. We need the ability to push changes out without having to change every repo. I added an inline comment in both pull-request.yml and release.yml explaining that additional-build-command must run after npm run build. If auro-actions ever reorders that, the comment should give the next person investigating a fighting chance to spot it.

test/check-bundled-imports.mjs
rmenner: Pulled findUnbundledImports + buildAllowlist into packages/utils/ and added 4 smoke tests. The CLI now imports from there, so a regex regression fails the tests instead of silently disarming the gate.

test/check-bundled-imports.mjs
rmenner: Verifier now walks the root package.json exports map and checks every ./components/*/dist/*.js it finds, future entry points are picked up automatically. Also split the error hint so missing files point at npm run build, unbundled imports point at EXTERNAL_PACKAGE_NAMES.

.github/workflows/pull-request.yml
rmenner: Added!

packages/config/src/internal.rollup.config.mjs
```diff
@@ -2,20 +2,38 @@
...
     entryFileNames: '[name].js'
   },
-  external: [
-    /node_modules\/lit/,
-    /node_modules\/lit-element/,
-    /node_modules\/lit-html/,
-    /node_modules\/@lit/,
-    /node_modules\/@lit-labs/,
-  ],
+  external: EXTERNAL_PACKAGE_NAMES.map((name) => new RegExp(`node_modules/${name}`)),
```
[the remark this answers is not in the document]
rmenner: Anchored the rollup external regex to a path boundary so it matches the verifier's allowlist semantics. Skipping the Windows part since rollup normalizes IDs to forward slashes and CI is Linux.

packages/utils/test/unbundledImports.test.js
```diff
@@ -0,0 +1,30 @@
+import { expect } from '@open-wc/testing';
+import { buildAllowlist, findUnbundledImports, init } from '../src/unbundledImports.js';
+
+const ALLOWED = buildAllowlist(['lit', 'lit-element', 'lit-html', '@lit', '@lit-labs']);
+
+before(async () => {
+  await init;
+});
```
chrisfalaska, line 8: es-module-lexer ships its parser as WASM. WTR runs in a real browser, so it should load fine, but the first CI run is the proof.

Confirm the first WTR run actually loads it -- if the WASM fetch hangs or 404s, await init will sit forever and the suite times out without a useful error.

If you see  flakes here, consider bumping the mocha timeout and/or adding a setTimeout race with a clear message.

chrisfalaska (changes requested):

packages/utils/test/unbundledImports.test.js
rmenner: Sure we can make adjustments once it's in.

chrisfalaska (approved):

PR #1508, opened 2026-06-23
v2 - formkit v6/address comments

jason-capsule42:

## Review Checklist:

<details>
<summary>
RC Checklist
</summary>

### Browsers

Android

- [ ] Chrome
- [ ] Firefox

 iOS

- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Desktop

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### Scenarios

- [ ] Next React
- [ ] SvelteKit

</details><br>

-- Auro Design System Team

## Summary by Sourcery

Refine combobox selection, validation, and masking behaviors and simplify date handling in base-input while updating related tests and utilities.

Bug Fixes:
- Ensure combobox trigger and bib inputs stay in sync via a shared helper, including suggestion-mode freeform values and filter behavior.
- Prevent stale menu selections from repopulating combobox values by clearing menu selection state consistently when values no longer match options.
- Avoid announcing undefined selections to screen readers and move focus to the clear button after a user selection for better accessibility.
- Fix caret positioning and focus management when opening the combobox popover or fullscreen dialog across desktop and mobile.
- Align combobox validity with its internal validation state, including credit card scenarios, and simplify reset behavior to clear validity consistently.
- Resolve date input masking edge cases by reconfiguring IMask safely when formats change and by excluding date from certain mask value paths to preserve correct validity.

Enhancements:
- Streamline base-input date handling by removing internal Date object sync fields and computing value/min/max objects directly from validated ISO strings.
- Clarify and tighten combobox option activation logic and keyboard navigation index reconciliation.
- Add debugging output for invalid date formatting in input utilities to aid troubleshooting.

Tests:
- Update combobox tests to validate against the component’s validity state, exercise keyboard-driven selection flows, and cover mobile versus desktop differences.
- Remove obsolete auro-input date object synchronization tests that relied on the removed internal setters.

### DIFF
--- a/components/combobox/src/auro-combobox.js
+++ b/components/combobox/src/auro-combobox.js
@@ -773,11 +773,8 @@ export class AuroCombobox extends AuroElement {
       option.setAttribute('aria-posinset', index + 1);
     });
 
-    if (this.value && this.input.value && !this.menu.value) {
-      if (this.behavior === 'suggestion' && this.menu.options && this.menu.options.some((opt) => opt.value === this.value)) {
-        this.setMenuValue(this.value);
-      }
-
+    if (this.input.value && this.menu.options && this.menu.options.some((opt) => opt.value === this.input.value)) {
+      this.setMenuValue(this.input.value);
       this.syncValuesAndStates();
     }
 
@@ -1328,8 +1325,6 @@ export class AuroCombobox extends AuroElement {
         });
       }
 
-      this.value = this.input.value;
-
       this.dispatchEvent(new CustomEvent('inputValue', { detail: { value: this.inputValue } }));
     } else if (event.target === this.input) {
 
@@ -1344,6 +1339,8 @@ export class AuroCombobox extends AuroElement {
         bibNativeInput.value = this.input.value || '';
       }
     }
+
+    this.value = this.input.value;
 
     // Ignore re-entrant input events caused by programmatic value sets.
     if (this._syncingBibValue || this._syncingDisplayValue) {
--- a/components/combobox/src/auro-combobox.js
+++ b/components/combobox/src/auro-combobox.js
@@ -147,6 +147,7 @@ export class AuroCombobox extends AuroElement {
     this.touched = false;
     this.validation = new AuroFormValidation();
     this.validity = undefined;
+    this._userTyped = false;
   }
 
   // This function is to define props used within the scope of this component
@@ -540,13 +541,6 @@ export class AuroCombobox extends AuroElement {
     }
     return this.input.value;
   }
-
-  // /**
-  //  * Sets the value of the input element within the combobox.
-  //  */
-  // set inputValue(value) {
-  //   this.input.value = value;
-  // }
 
   /**
    * Checks if the element is valid.
@@ -1177,9 +1171,7 @@ export class AuroCombobox extends AuroElement {
       // Hide dropdown on selection (except during slot changes)
       if (evt.detail && evt.detail.source !== 'slotchange') {
         // do not close while typing in suggestion mode with no value selected, to allow freeform input
-        if (this.menu.value || this.behavior !== 'suggestion') {
-          this.hideBib();
-        }
+        this.hideBib();
 
         // Move focus to the clear button when the user makes a selection.
         if (!isEcho && this.menu.value !== undefined) {
@@ -1205,10 +1197,6 @@ export class AuroCombobox extends AuroElement {
       // only — fresh user selections take the existing hideBib path.
       if (isEcho && this.menu.optionSelected) {
         this._programmaticFilterRefresh = true;
-        this.handleMenuOptions();
-        setTimeout(() => {
-          this._programmaticFilterRefresh = false;
-        }, 0);
       }
     });
 
@@ -1331,48 +1319,36 @@ export class AuroCombobox extends AuroElement {
       if (this._syncingDisplayValue) {
         return;
       }
-      this._syncingBibValue = true;
-      this.input.value = this.inputInBib.value;
-      this.input.updateComplete.then(() => {
-        this._syncingBibValue = false;
-      });
 
-      // Run filtering inline — the re-entrant event won't reach this code.
-      this.menu.matchWord = normalizeFilterValue(this.inputInBib.value);
-      this.optionActive = null;
-
-      // In suggestion mode, keep the combobox value in sync with the typed
-      // text so that freeform values are captured (mirroring the non-fullscreen
-      // path). Clear the selection when the input is emptied.
-      if (this.behavior === 'suggestion') {
-        this.value = this.inputInBib.value || undefined;
+      if (this.input.value !== this.inputInBib.value) {
+        this._syncingBibValue = true;
+        this.input.value = this.inputInBib.value;
+        this.input.updateComplete.then(() => {
+          this._syncingBibValue = false;
+        });
       }
 
-      this.handleMenuOptions();
+      this.value = this.input.value;
+
       this.dispatchEvent(new CustomEvent('inputValue', { detail: { value: this.inputValue } }));
-      return;
+    } else if (event.target === this.input) {
+
+      // Also sync the native input immediately so keystrokes arriving
+      // before Lit's async update cycle (e.g. rapid Backspaces during a
+      // fullscreen transition) operate on the correct content.
+      // Skip the next programmatic input event to prevent the patched setter
+      // from re-entering handleInputValueChange via the bib path.
+      const bibNativeInput = this.inputInBib.inputElement;
+      if (bibNativeInput && bibNativeInput.value !== this.input.value) {
+        this.inputInBib.skipNextProgrammaticInputEvent = true;
+        bibNativeInput.value = this.input.value || '';
+      }
     }
 
     // Ignore re-entrant input events caused by programmatic value sets.
     if (this._syncingBibValue || this._syncingDisplayValue) {
       return;
     }
-
-    this.inputInBib.value = this.input.value;
-
-    // Also sync the native input immediately so keystrokes arriving
-    // before Lit's async update cycle (e.g. rapid Backspaces during a
-    // fullscreen transition) operate on the correct content.
-    // Skip the next programmatic input event to prevent the patched setter
-    // from re-entering handleInputValueChange via the bib path.
-    const bibNativeInput = this.inputInBib.inputElement;
-    if (bibNativeInput && bibNativeInput.value !== this.input.value) {
-      this.inputInBib.skipNextProgrammaticInputEvent = true;
-      bibNativeInput.value = this.input.value || '';
-    }
-
-    this.menu.matchWord = normalizeFilterValue(this.input.value);
-    this.optionActive = null;
 
     if (this.behavior === 'suggestion') {
       this.value = this.input.value;
@@ -1381,7 +1357,6 @@ export class AuroCombobox extends AuroElement {
     if (!this.input.value && !this._clearing) {
       this.clear();
     }
-    this.handleMenuOptions();
 
     // Validate only if the value was set programmatically (not during user
     // interaction). In fullscreen dialog mode, componentHasFocus returns false
@@ -1391,34 +1366,37 @@ export class AuroCombobox extends AuroElement {
       this.validate();
     }
 
-    if (this.input.value && this.input.value.length === 0) {
-      // Hide menu if value is empty, otherwise show if there are available suggestions
-      this.hideBib();
-    } else if (this.menu.loading) {
-      // if input has value but menu is loading, show bib immediately
-      this.showBib();
-    } else if (this.availableOptions.length === 0 && !this.dropdown.isBibFullscreen) {
-      // Force dropdown bib to hide if input value has no matching suggestions
-      this.hideBib();
-    }
-
-    // iOS virtual keyboard retention: when in fullscreen mode, ensure the
-    // dialog opens and the bib input is focused synchronously within the
-    // input event (user gesture) chain. Without this, Lit's async update
-    // cycle delays showModal() past the user activation window, causing
-    // iOS Safari to dismiss the virtual keyboard when the fullscreen
-    // dialog opens — the user then has to tap the input again to resume
-    // typing.
-    if (this.dropdown.isBibFullscreen && this.input.value && this.input.value.length > 0) {
-      if (!this.dropdown.isPopoverVisible) {
-        this.showBib();
-      }
-      if (this.dropdown.isPopoverVisible) {
-        this.setInputFocus();
-      }
-    }
-
     this.dispatchEvent(new CustomEvent('inputValue', { detail: { value: this.inputValue } }));
+  }
+
+  /**
+   * Handles input value changes originating from the trigger input.
+   * Refreshes menu options and filtering, delegates to handleInputValueChange
+   * for value synchronization, and manages fullscreen bib focus.
+   * @private
+   * @param {Event} event - The input event from the trigger input element.
+   * @returns {void}
+   */
+  handleTriggerInputValueChange(event) {
+    this.handleMenuOptions();
+
+    // Run filtering inline — the re-entrant event won't reach this code.
+    this.menu.matchWord = normalizeFilterValue(this.inputInBib.value);
+    this.optionActive = null;
+
+    if (this.value === this.input.value) {
+      return;
+    }
+
+    this.handleInputValueChange(event);
+
+    if (this.dropdown.isBibFullscreen && this.input.value && this.input.value.length > 0 && this.dropdown.isPopoverVisible) {
+      this.setInputFocus();
+    }
+
+    if (this._programmaticFilterRefresh) {
+      this._programmaticFilterRefresh = false;
+    }
   }
 
   /**
@@ -1597,43 +1575,8 @@ export class AuroCombobox extends AuroElement {
         this.input.value = this.value;
       }
 
-      // Sync menu.value only when an option actually matches, or when filter
-      // mode needs setMenuValue to dispatch auroMenu-selectValueFailure (the
-      // listener at line 1206 clears combobox.value, mirroring select's
-      // pattern). Otherwise sync the input display for freeform values.
-      if (this.menu.options && this.menu.options.length > 0) {
-        if (this.menu.options.some((opt) => opt.value === this.value) || this.behavior === 'filter') {
-          this.setMenuValue(this.value);
-        } else {
-          // Clear menu.value AND menu.optionSelected together. Clearing only
-          // menu.value leaves the previously-selected option element pinned
-          // as menu.optionSelected; a later auroMenu-selectedOption event
-          // would then write its stale .value back into combobox.value.
-          if (this.menu.value || this.menu.optionSelected) {
-            this.menu.clearSelection();
-          }
-          // Suggestion-mode freeform value: sync the trigger + bib to show it,
-          // then refresh the filter once the inputs flush (handleInputValueChange
-          // bailed on _syncingDisplayValue).
-          const syncPromise = this.syncInputValuesAcrossTriggerAndBib(this.value || '');
-          if (syncPromise) {
-            syncPromise.then(() => {
-              this._programmaticFilterRefresh = true;
-              this.handleMenuOptions();
-              setTimeout(() => {
-                this._programmaticFilterRefresh = false;
-              }, 0);
-            });
-          }
-        }
-      }
       if (!this.value) {
         this.clear();
-      }
-      if (this.value && !this.componentHasFocus) {
-        // If the value got set programmatically make sure we hide the bib
-        // when input is not taking the focus (input can be in dropdown.trigger or in bibtemplate)
-        this.hideBib();
       }
 
       // Sync the input and match word, but don't directly set menu.value again
@@ -1670,15 +1613,27 @@ export class AuroCombobox extends AuroElement {
       // from a programmatic filter refresh after a value swap — the user
       // didn't interact, so we shouldn't auto-open the bib (especially for
       // the no-match path where the existing condition unconditionally fires).
-      if (this._programmaticFilterRefresh) {
-        this._programmaticFilterRefresh = false;
-      } else if ((this.availableOptions.length > 0 && (this.componentHasFocus || this.dropdownOpen)) || (this.menu && this.menu.loading) || (this.availableOptions.length === 0 && this.noMatchOption)) {
-        this.showBib();
+      if ((this.menu && !this._programmaticFilterRefresh)) {
+        if (
+          this.availableOptions.length > 0 ||
+          this.menu.loading ||
+          this.noMatchOption
+        ) {
+          if (!this.dropdownOpen && this._userTyped) {
+            this.showBib('availableOptions');
+            this._userTyped = false;
+          }
+        }
+
         if (!this.availableOptions.includes(this.menu.optionActive)) {
           this.activateFirstEnabledAvailableOption();
         }
-      } else if (this.dropdown && this.dropdown.isPopoverVisible) {
+      } else if (!this.dropdown.isBibFullscreen) {
         this.hideBib();
+      }
+
+      if (this._programmaticFilterRefresh) {
+        this._programmaticFilterRefresh = false;
       }
     }
 
@@ -1822,7 +1777,7 @@ export class AuroCombobox extends AuroElement {
           shape="${this.shape}"
           size="${this.size}">
             <${this.inputTag}
-              @input="${this.handleInputValueChange}"
+              @input="${this.handleTriggerInputValueChange}"
               appearance="${this.onDark ? 'inverse' : this.appearance}"
               .a11yActivedescendant="${this.dropdownOpen && this.optionActive ? this.optionActive.id : undefined}"
               .a11yExpanded="${this.triggerExpandedState}"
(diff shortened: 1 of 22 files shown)


components/combobox/src/comboboxKeyboardStrategy.js
```diff
@@ -101,13 +100,8 @@ export const comboboxKeyboardStrategy = {
   },
 
   Enter(component, evt, ctx) {
-    if (isClearBtnFocused(ctx)) {
```
jason-capsule42, line 104: Why did we remove this? Now the combobox will make a selection with the enter key when focus is on the clear button which should be clearing the selected value, not making a new selection.

components/combobox/src/comboboxKeyboardStrategy.js
```diff
@@ -116,14 +110,12 @@ export const comboboxKeyboardStrategy = {
       }
 
       evt.preventDefault();
-      evt.stopPropagation();
```
jason-capsule42, line 119: Why would we want the form to pick up an Enter key event when activating the Clear button. Also, Forms should not do Enter key based submission for combobox elements any how. I don't think we should propagate this event.

packages/form-validation/src/validation.js
```diff
@@ -321,15 +321,17 @@ export default class AuroFormValidation {
...
+      // If there is a second input in the elem and that value is undefined or an empty string set hasValue to false.
+      // Skip for combobox: its second auro-input is the fullscreen-bib mirror of the same value, not an independent
+      // field (datepicker is the intended consumer — start/end are independently required).
+      if (this.auroInputElements?.length === 2 && !isCombobox) {
+        if (!this.auroInputElements[1].value || this.auroInputElements[1].value.length === 0) {
           hasValue = false;
         }
       }
 
-      const isCombobox = this.runtimeUtils.elementMatch(elem, 'auro-combobox');
```
jason-capsule42, line 331: What was this originally doing that prompted the removal?

rmenner: This was just moved a few lines up so it could be reused.

components/combobox/src/comboboxKeyboardStrategy.js
rmenner: Fixed it. Propagation should always be stopped.

components/combobox/src/comboboxKeyboardStrategy.js
rmenner: Didn't mean to change it. Restored the early return so Enter on the focused clear button activates the button (clearing the value) instead of making a menu selection.

components/combobox/test/auro-combobox.test.js
```diff
@@ -3486,35 +3500,25 @@ function runFullTest(mobileView) {
...
-
         await sendKeys({ press: 'o' });
-
         await elementUpdated(el);
-
         await sendKeys({ press: 'Tab' });
+        await elementUpdated(el);
 
-        await expect(el.value).to.be.equal(options[1].textContent);
+        await expect(el.value === options[1].textContent);
```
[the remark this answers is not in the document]
rmenner: `await expect(el.value === options[1].textContent);` was a no-op. Changed to a real Chai assertion

components/combobox/src/auro-combobox.js
```diff
@@ -749,6 +718,48 @@ export class AuroCombobox extends AuroElement {
...
+   *   or null when no sync was needed (so callers can skip post-flush work).
+   * @private
+   */
+  syncInputValuesAcrossTriggerAndBib(nextValue) {
+    // Only set the flag when there's an actual write to suppress —
+    // syncValuesAndStates re-enters here during typing when both inputs
+    // already match, and a no-op flag flip would make the bib branch's
+    // bail eat legitimate user-input events.
+    const triggerNeedsSync = this.input.value !== nextValue;
+    const bibNeedsSync = this.inputInBib.value !== nextValue;
```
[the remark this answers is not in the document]
rmenner: `bibNeedsSync` now short-circuits when `this.inputInBib` is falsy

components/combobox/src/auro-combobox.js
```diff
@@ -541,13 +542,6 @@ export class AuroCombobox extends AuroElement {
     return this.input.value;
   }
 
-  // /**
-  //  * Sets the value of the input element within the combobox.
-  //  */
-  // set inputValue(value) {
-  //   this.input.value = value;
-  // }
```
jason-capsule42, line 549: Remove dead code

rmenner: that is removed

components/combobox/src/auro-combobox.js
```diff
@@ -1565,6 +1542,13 @@ export class AuroCombobox extends AuroElement {
       this.optionSelected = undefined;
       this.value = undefined;
 
+      // clear the displayValue in the trigger if displayValue slot content is present
+      const displayValueInTrigger = this.input.querySelector('[slot="displayValue"]');
+
+      if (displayValueInTrigger) {
+        displayValueInTrigger.remove();
```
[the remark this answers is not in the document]
rmenner: fixed

components/combobox/src/auro-combobox.js
```diff
@@ -1326,57 +1321,44 @@ export class AuroCombobox extends AuroElement {
...
+      if (this.input.value !== this.inputInBib.value) {
+        this._syncingBibValue = true;
+        this.input.value = this.inputInBib.value;
+        this.input.updateComplete.then(() => {
+          this._syncingBibValue = false;
+        });
       }
 
-      this.handleMenuOptions();
       this.dispatchEvent(new CustomEvent('inputValue', { detail: { value: this.inputValue } }));
```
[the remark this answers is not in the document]
rmenner: I added a comment explaining what happens in the lit lifecycle that makes this work as is

components/combobox/src/auro-combobox.js
```diff
@@ -1588,81 +1572,21 @@ export class AuroCombobox extends AuroElement {
   }
 
   updated(changedProperties) {
-    // After the component is ready, send direct value changes to auro-menu.
+    // After the component is ready, propogate direct changes down to child components.
```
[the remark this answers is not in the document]
rmenner: fixed

components/combobox/test/auro-combobox.test.js
```diff
@@ -1915,18 +1965,21 @@ function runFullTest(mobileView) {
 
     describe('showBib', () => {
       it('should show the dropdown when input has matching options', async () => {
-        const el = await defaultFixture(mobileView);
+        const el = await defaultFixture();
```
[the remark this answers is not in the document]
rmenner: fixed

chrisfalaska:
# Adversarial-derived review

## P0 — Blockers (fix before merge)

### ** NO CHANGE (JASON)** P0-1. Focus moves to clear button after selection (a11y + UX regression)

**RESOLUTION** Current behavior is as intended per A11Y review internally.

**Sources:** a11y (HIGH), tests (HIGH — untested), correctness, code-quality.
**Location:** `components/combobox/src/auro-combobox.js:1173-1176`; doc at `components/combobox/docs/partials/keyboard-behavior/keyEvents.md:148`.

After Enter/click selection, focus is moved to the clear button. Issues:
- Violates the WAI-ARIA editable-combobox APG pattern (focus must return to the textbox so the user can keep typing).
- Causes VoiceOver double-talk: `"clear, button"` collides with the polite live-region `"<value>, selected"` announcement at `:1186`.
- Breaks Tab predictability — Shift+Tab from a freshly selected combobox now lands on the clear button instead of the previous field.
- `keyEvents.md:148` documents this for Tab but does not disambiguate the "no value selected" case.
- **No regression test** asserts `el.shadowRoot.activeElement === clearBtn` after selection.

**Action:** Decide on intended behavior. If keeping: gate it (Enter only, not click; suppress live-region announcement on same tick) and add a regression test. If reverting: focus stays on the input.

---

### **RESOLVED** P0-2. eef7900d4's stale `menu.optionSelected` write-back fix appears reverted

RESOLUTION: combobox now uses menus clear function to properly reset it.

**Sources:** correctness (HIGH), tests (HIGH).
**Location:** `components/combobox/src/auro-combobox.js:1580`.

The rewrite of `updated()` now does `this.menu.value = undefined` but no longer clears `menu.optionSelected`. The original eef7900d4 fix is gone.

**Repro:** Set `combobox.value` programmatically to an unmatched string while a prior option is selected → the next `auroMenu-selectedOption` echo writes the stale `optionSelected.value` back into `combobox.value`. This is exactly the bug eef7900d4 was meant to kill.

**Action:** Restore `menu.clearSelection()` (or equivalent) inside the unmatched-value branch. Add the regression test that was missing originally.

---

### P0-3. **RESOLVED** Paste / IME / dead-key input does not open the bib

RESOLUTION: Fix applied. The previous keydown.key.length === 1 gate in the keyboard strategy missed paste (no keydown), IME composition (key='Process'), and dead-key composition (multi-char keys). The flag is now flipped on the input event, which is the canonical signal that fires for all user input methods including paste, compositionend, and drop.

**Sources:** a11y (MED), correctness (MED).
**Location:** `components/combobox/src/comboboxKeyboardStrategy.js:172-174`.

`_userTyped` only flips when `evt.key.length === 1`. This misses:
- Paste events (no keydown).
- IME composition (`evt.key === 'Process'`).
- Dead-key composition (multi-char keys).

Result: `updated('availableOptions')` never sees `_userTyped === true` for these inputs and never calls `showBib()` (`auro-combobox.js:1631`). Paste-to-search and CJK input are broken.

**Action:** Listen to the host `input` event or `compositionend` to flip `_userTyped`, rather than gating purely on `keydown.key.length`.

---

## P1 — High (should fix before merge)

### **RESOLVED** P1-1. Live region double-announces on every arrow keypress

RESOLUTION: Fix applied. The manual live-region mirror in the auroMenu-activatedOption listener now only fires when the bib is fullscreen, so popover mode lets the native aria-activedescendant mechanism do its job without competing announcements.

**Source:** a11y (HIGH).
**Location:** `components/combobox/src/auro-combobox.js:1223-1228`.

`auroMenu-activatedOption` writes `"<text>, selected/not selected, X of Y"` to the polite live region on every ArrowUp/Down. The trigger input also exposes `aria-activedescendant` (the WAI-ARIA mechanism for option announcement). VoiceOver announces both → flooded polite queue on key-repeat.

**Action:** Gate the manual announcement to the fullscreen-dialog code path only (where shadow-root boundaries actually break native AT). Remove for non-fullscreen.

---

### **RESOLVED** P1-2. `aria-expanded="true"` may never reach assistive tech
**Source:** a11y (HIGH).
**Location:** `components/combobox/src/auro-combobox.js:858-866`.

RESOLUTION: Fix applied. The 150ms deferral on aria-expanded "true" is gone — the trigger input now binds directly to dropdownOpen so AT sees the state transition synchronously with the actual open.

`triggerExpandedState` is set to `true` via a 150 ms timer to dodge VoiceOver character-echo collisions. If the user opens and selects in under 150 ms (common with typeahead), the dropdown closes before the timer fires and AT only ever sees "collapsed → collapsed".

**Action:** Remove the deferral; rely on activedescendant + announcement gating to handle echo collisions instead of lying about expanded state.

---

### **RESOLVED** P1-3. `getOptionLabel` clones the DOM on every keystroke

RESOLUTION: Fix applied. getOptionLabel now walks direct childNodes and filters the displayValue slot in-place — no DOM allocation per call.

**Source:** perf (HIGH).
**Location:** `components/combobox/src/auro-combobox.js:54-64`.

`cloneNode(true)` + `querySelector` + regex on textContent. Called from `syncValuesAndStates` (every input event) and the option-selected listener. The equivalent select-side code is a single `innerHTML` write.

**Action:** Replace with a `for-of` over `option.childNodes` filtering out the `displayValue` slot — no DOM allocation, same text result. Or prefer `dataset.label` when present.

---

### **RESOLVED** P1-4. `keyEvents.md` Meta/Alt+Arrow rows deleted while feature still ships

RESOLUTION: Removed the Meta/Alt/Ctrl+Arrow shortcut — Home/End remain as the documented way to jump to first/last enabled options.

**Source:** docs/tests (HIGH).
**Locations:**
- Implementation: `components/combobox/src/comboboxKeyboardStrategy.js:60, 83`.
- Test: `apps/shared/combobox-interaction.suite.ts:140-181`.
- Doc: `components/combobox/docs/partials/keyboard-behavior/keyEvents.md` (rows removed).

The Meta/Alt+ArrowDown/Up "jump to first/last enabled option" feature is implemented and tested but no longer documented. Commit `7f32b98f3` claims to update these notes — contradicted by the actual diff.

**Action:** Restore the rows. Users have no way to discover a working, tested feature.

---

## P2 — Medium (fix soon)

### **RESOLVED** P2-1. Double filter write per keystroke

RESOLUTION: Consolidated to a single menu.matchWord write per keystroke inside handleMenuOptions.

**Source:** perf (MED).
**Location:** `components/combobox/src/auro-combobox.js:1383-1386`.

`handleTriggerInputValueChange` calls `handleMenuOptions` then re-sets `menu.matchWord`, even though `updateFilter` (called inside `syncValuesAndStates`) already did. Two filter writes per keystroke.

---

### **RESOLVED** P2-2. `syncInputValuesAcrossTriggerAndBib` errors swallowed

RESOLUTION: Fix applied. An imask throw can no longer strand _syncingDisplayValue in the true state.

**Source:** correctness (MED).
**Location:** `components/combobox/src/auro-combobox.js:745, 1583`.

The returned promise is never awaited or caught. If `maskInstance.updateValue()` throws (commit `d1857401c` says imask can throw on credit-card format change), `_syncingDisplayValue` stays `true` forever, breaking all subsequent inputs.

**Action:** `await` the call inside a `try/finally` that always clears `_syncingDisplayValue`.

---

### **RESOLVED** P2-3. Tests assert on private state instead of observable behavior

RESOLUTION: Reworked the cases that had clean observable equivalents, kept the ones that don't with explicit "regression guard" rationale.

**Source:** tests (MED).
**Locations:** `components/combobox/test/auro-combobox.test.js:61, 756, 2142, 2332, 2417, 2433, 2454`.

Examples: spying on `performUpdate`, `Object.defineProperty(el.dropdown, 'isBibFullscreen', …)`, asserting on `_inFullscreenTransition`, `isHiddenWhileLoading`. Some warranted (regression guards) but several should be DOM/ARIA assertions.

---

### **RESOLVED** P2-4. Skipped / known-flaky tests with TODO comments

**RESOLUTION** Jason cleaned up tests

**Source:** tests (MED).
**Locations:** `components/combobox/test/auro-combobox.test.js:105, 634, 672, 702, 3325`.

Comments like `// IS THIS FAILING DUE TO A BUG?` and `// TODO: revisit` without follow-up tickets hide gaps.

**Action:** File issues or remove. Either the bug is real and needs fixing, or the test needs to be deleted/rewritten.

---

### **RESOLVED** P2-5. `why-combobox.md` claims `aria-owns`, source uses `aria-controls`

RESOLUTION: corrected the inaccurate ARIA list.

**Source:** docs (MED).
**Location:** `components/combobox/docs/pages/why-combobox.md:12`.

---

### **RESOLVED** P2-6. No-match programmatic-value silently closes bib with no SR feedback

RESOLUTION: Fix applied at both code paths the silent-close can take.

**Source:** a11y (MED).
**Locations:** `components/combobox/src/auro-combobox.js:648, 1208`.

If a consumer sets `value` to a string with no matching option, the bib closes silently. `auroMenu-selectValueFailure` clears state but does not announce. Screen reader users get no signal that their request was rejected.

---

### **RESOLVED** P2-7. Four overlapping timers per open cycle, no consolidated cancellation

RESOLUTION: All four setTimeout call sites now go through a single tracker that disconnectedCallback drains. The _expandedTimeout the reviewer cited at :861 was already gone from the previous fix.

**Source:** perf (MED).
**Locations:**
- `_expandedTimeout` at `:861`.
- `auroDropdown-strategy-change` setTimeout at `:972`.
- `setMenuValue` backup clear setTimeout at `:1510`.
- Announcement timer at `:1185`.

None are cancelled in `disconnectedCallback`.

---

### **RESOLVED** P2-8. Missing regression tests for recent bug fixes

RESOLUTION: Three regression tests added. The eef7900d4 "menu.optionSelected cleared" test was already in the suite from a previous turn — no duplicate needed.

**Source:** tests (HIGH, but no immediate user impact).
**Location:** `components/combobox/test/auro-combobox.test.js`.

Each of these landed without a test:
- Caret restoration on reopen (`f876902e2`).
- Focus-to-clear-button selection (`d9b8eb2c6`, `4b90416db`) — see P0-1.
- `menu.optionSelected` cleared (`eef7900d4`) — see P0-2.
- `bibNeedsSync` simplification (`0146fd859`) — only indirectly covered.

---

## **RESOLVED** P3 — Low / nits

RESOLUTION:
```
Resolved with edits:

[apps/shared/combobox-interaction.suite.ts:372-385](vscode-webview://03fttbe88tddn2dic8k2dne3e7kl0qdf5258kbfps8opbfgrc87p/apps/shared/combobox-interaction.suite.ts#L372-L385) — P3-9. Tightened the click-outside assertion from not.toBe('Oranges') to toBe('a') (the typed text in suggestion mode). The prior form would have passed silently if 'Apples' was selected; the strict form catches any auto-selection regression.

[components/combobox/test/auro-combobox.test.js:4361-4444](vscode-webview://03fttbe88tddn2dic8k2dne3e7kl0qdf5258kbfps8opbfgrc87p/components/combobox/test/auro-combobox.test.js#L4361-L4444) — P3-7. Added two regression tests for commit 4271eed64's disabled-option skipping:

End should skip a disabled last option and activate the prior enabled one — symmetric to the existing Home test.
ArrowUp wrap skips a disabled trailing option when wrapping from the first — covers the menu's wrap-around path.
[components/combobox/test/auro-combobox.test.js:3045-3057](vscode-webview://03fttbe88tddn2dic8k2dne3e7kl0qdf5258kbfps8opbfgrc87p/components/combobox/test/auro-combobox.test.js#L3045-L3057) — P3-3. Replaced the setTimeout(resolve, 2200) blocking wait with waitUntil(() => liveRegion.textContent === '', …, { timeout: 3000 }). Test now returns as soon as the cleanup timer fires (~1300ms in practice), removing the worst flake-prone wait the reviewer flagged.

Verified no action needed:

P3-1 (propogate typo at :1575): No such typo anywhere in [components/combobox/src/auro-combobox.js](vscode-webview://03fttbe88tddn2dic8k2dne3e7kl0qdf5258kbfps8opbfgrc87p/components/combobox/src/auro-combobox.js). Likely fixed already; current line 1575 reads this.menu.value = undefined;.
P3-5 (duplicate "Hide the trigger from AT" comments at :912 / :917): Only one "Hide the trigger from assistive technology" comment exists now (line 936). Line numbers have shifted significantly through prior edits.
P3-2 (typedValue dead): NOT dead — referenced in [components/combobox/apiExamples/dynamic-menu.js](vscode-webview://03fttbe88tddn2dic8k2dne3e7kl0qdf5258kbfps8opbfgrc87p/components/combobox/apiExamples/dynamic-menu.js), the Svelte + React framework apps, and the [docs/api.md](vscode-webview://03fttbe88tddn2dic8k2dne3e7kl0qdf5258kbfps8opbfgrc87p/components/combobox/docs/api.md) public API. Documented WRITE-only property. Removal would break consumers.
P3-4 (availableOptions.hasChanged rarely fires): The custom hasChanged does content-equality comparison, so it short-circuits when an incremental keystroke yields the same option set (e.g. typing past the first unique match). Narrow but not dead — kept.
Out of scope as noted:

P3-6 (WHAT vs WHY comments at multiple line ranges): Skipped. Sweeping comment rewrite across non-overlapping sections would balloon the diff; happy to take a specific pass if you want to point at one or two passages.
P3-8 (customize.md missing persistInput / noFilter examples): Skipped. Docs additions need example content I don't have here; happy to draft if you give me the desired example shape.
P3-3 (the broader 30-occurrence cleanup of setTimeout(resolve, 0) / double-rAF): Only addressed the 2200ms outlier the reviewer specifically called out. The 0ms / rAF waits are mostly anchor points after specific async events and converting each to waitUntil requires per-case observable selection — a separate sweep if desired.
```

- **P3-1.** Typo: `propogate` at `components/combobox/src/auro-combobox.js:1575`.
- **P3-2.** `typedValue` property at `:457` appears effectively dead post-refactor.
- **P3-3.** 30 occurrences of `setTimeout(resolve, 0)` / double-rAF in tests including a **2200 ms** wait at `auro-combobox.test.js:2785` — will flake under CI load. Prefer `waitUntil` against observable state.
- **P3-4.** `availableOptions.hasChanged` optimization at `:195-206` rarely fires — `handleMenuOptions` allocates a new array each call.
- **P3-5.** Duplicate "Hide the trigger from assistive technology" comments at `:912` and `:917`.
- **P3-6.** Many new comments narrate WHAT, not WHY (e.g. `:1311-1331`, `:1431-1438`, `:1620-1624`).
- **P3-7.** Disabled-option skipping (`4271eed64`) is covered for Home (`auro-combobox.test.js:3913`) but not for End or ArrowUp wrap.
- **P3-8.** `customize.md` adds no new examples for `persistInput` / `noFilter` despite richer behavior in the new branch.
- **P3-9.** `combobox-interaction.suite.ts:415-426` "click outside" asserts `value).not.toBe('Oranges')` — passes even if click silently selects `'Apples'`. Tighten to `.toBeUndefined()`.

---

## P4 — Follow-up (next PR, not a blocker)

### **DEFERRED** P4-1. State machine refactor for re-entrancy guards
**Source:** code-quality (HIGH), correctness (multiple bugs traced here).

8 ad-hoc flags coordinate event re-entrancy: `_pendingMenuValueSync`, `_programmaticFilterRefresh`, `_syncingBibValue`, `_syncingDisplayValue`, `_clearing`, `_userTyped`, `_inFullscreenTransition`, `skipNextProgrammaticInputEvent`. Each is commented to explain why the others aren't enough. `setMenuValue` even has a `setTimeout(0)` "backup clear" admitting the primary mechanism is unreliable (`:1510`).

The +522 line file growth is primarily plumbing for these flags — perhaps ~150 lines are real fixes, the rest is special-case accretion around `setMenuValue` / `updated('value')` / `handleInputValueChange`.

**Recommendation:** Do not ship more features into this layer without a small state-machine refactor. Probable shape: a `_pendingSource` enum (`'user-type' | 'programmatic' | 'menu-select' | 'clear'`) replaces most boolean flags.

### **DEFERRED** P4-2. `updated('value')` → `clear()` → `this.value = undefined` re-entrancy
**Source:** correctness, code-quality.
**Location:** `components/combobox/src/auro-combobox.js:1587, 1543`.

`updated()` calls `clear()` when value falsy; `clear()` sets `this.value = undefined`, re-entering `updated`. The `_clearing` guard catches it but the dependency is implicit. Compare with select's equivalent which calls `setMenuValue` without re-clearing. Would be addressed naturally by P4-1.

---

## Preserve — these changes look good

- `auroFormElement-validated` retarget guard at `:1436` — fixes a real composed-event bug.
- `focusout` guard against `_inFullscreenTransition` and `dropdownOpen` at `:1261` — replaces a setTimeout-based hack.
- `reconcileMenuIndex` in `comboboxKeyboardStrategy.js:38-45` — cleanly fixes `_index < 0` after async `clearSelection`.
- `syncInputValuesAcrossTriggerAndBib` consolidates trigger/bib write + imask refresh with `Promise.all` — good factoring.
- `aria-setsize` / `aria-posinset` on visible options at `:772-773` — compensates for shadow-DOM-broken listbox semantics.
- `bibDialogLabel` from `label` slot at `:950` — names the fullscreen dialog correctly.
- `trigger.inert = true` in fullscreen at `:913` — hides trigger from AT behind the modal.
- Cross-shadow `ariaActiveDescendantElement` wiring intact (`base-input.js:1116-1119`, `auro-combobox.js:1217, 870`).
- `getOptionLabel` strips the `displayValue` slot before assigning to `input.value` at `:54` — prevents SR from reading hidden display markup as part of the textbox value.
- `Escape` correctly returns focus to trigger in modal mode (`comboboxKeyboardStrategy.js:137`).
- `Enter` `stopPropagation` lifted to the top of the handler (`comboboxKeyboardStrategy.js:105`) — correctly prevents form submission across all sub-element focus cases.
- Performance regression-budget test at `auro-combobox.test.js:736-788` — clever guard against future cascade regressions.
- `combobox-interaction.suite.ts` uses real Playwright keyboard events and `expect.poll` against observable state. No private-state pokes. Right shape for the interaction layer.
- A11y suite at `auro-combobox.test.js:2733-3010` covers live region clear, fullscreen routing, cross-shadow activedescendant, `aria-expanded`, `aria-controls`. Solid baseline.
- New `why-combobox.md` and `css-only.md` are useful additions.
- Deleted `upstream.md`, `figma.md`, `design-notes.md` were boilerplate / 2022 changelog — safe to drop, no institutional knowledge lost.

components/input/src/utilities.js
```diff
@@ -275,6 +275,7 @@ export class AuroInputUtilities {
     const maskOptions = this.getMaskOptions('date', normalizedFormat);
 
     if (!(valueObject instanceof Date) || Number.isNaN(valueObject.getTime()) || !maskOptions || typeof maskOptions.format !== 'function') {
+      console.debug('Invalid date object or mask options for formatting', { valueObject, maskOptions }); // eslint-disable-line no-console
       return undefined;
     }
```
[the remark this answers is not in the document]
jason-capsule42: This console debug message is intentional and needs to remain.

rmenner (approved):
Approved (note: skipping some tests to work on later)

PR #1510, opened 2026-06-25
Jbaker/resolve datepicker review

jason-capsule42:

## Review Checklist:

<details>
<summary>
RC Checklist
</summary>

### Browsers

Android

- [ ] Chrome
- [ ] Firefox

 iOS

- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Desktop

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### Scenarios

- [ ] Next React
- [ ] SvelteKit

</details><br>

-- Auro Design System Team

## Summary by Sourcery

Strengthen auro-datepicker and auro-input timezone safety, blackout-date handling, and accessibility behavior, while adding regression tests and improving listener lifecycle management.

Bug Fixes:
- Ensure datepicker and date input ISO↔Date conversions are timezone-safe so displayed dates and value objects remain consistent across time zones.
- Fix range-preview behavior so hovering the departure date provides a single-day round-trip preview and correct ARIA range-position labels.
- Correct out-of-range calendar cells to use disabled buttons without redundant aria-hidden and ensure today cells expose aria-current="date".
- Prevent announceSelection from hanging or failing when the live region is late to mount or permanently unavailable.
- Resolve focus handling and keyboard navigation edge cases around month boundaries by waiting for calendar renders and picking the nearest focusable cell.
- Avoid stale event listeners by introducing an AbortController-based listener lifecycle and reconnect logic for auro-datepicker.
- Deduplicate invalid-locale fallback warnings so unsupported locale messages log once per tag and once per page class.

Enhancements:
- Introduce memoized blackout date handling shared between calendar and cells, including deprecation signaling for legacy disabledDays and validation of overflow ISO strings.
- Add explicit aria-roledescription="calendar" to the calendar grid and refine ARIA attributes on cells for clearer screen-reader output.
- Document default values for accessibility-related strings and new range end preview label in the API and localization docs.
- Improve calendar-month render coordination via a shared _afterMonthRender helper to make cross-month keyboard nav and focus restoration more robust.

Build:
- Run datepicker and input test suites under multiple time zones (PST, EST, HST) using cross-env, and narrow published demo artifacts in the package files list.

Tests:
- Add regression tests covering timezone-safe date conversions, blackout set memoization and deprecation warnings, range preview visual behavior, listener abort/reconnect flows, invalid-locale log deduping, live-region announcement retries, and calendar aria attributes.


jason-capsule42:
@sourcery-ai review

jason-capsule42:
@sourcery-ai review

jason-capsule42:
@sourcery-ai review

components/datepicker/src/auro-calendar.js
```diff
@@ -1588,7 +1708,7 @@ export class AuroCalendar extends RangeDatepicker {
           </button>
           ` : undefined}
         </div>
-        <div id="calendarGrid" class="calendars" role="group" tabindex="0" aria-label="${this.datepicker?.calendarGridLabel}" @keydown="${this.handleGridKeyDown}" @focusin="${this.handleGridFocusIn}" @focusout="${this.handleGridFocusOut}" @calendar-month-mouseleave="${this.clearRangePreview}">
+        <div id="calendarGrid" class="calendars" role="group" aria-roledescription="calendar" tabindex="0" aria-label="${this.datepicker?.calendarGridLabel}" @keydown="${this.handleGridKeyDown}" @focusin="${this.handleGridFocusIn}" @focusout="${this.handleGridFocusOut}" @calendar-month-mouseleave="${this.clearRangePreview}">
```
chrisfalaska, line 1725: The calendar container uses `role="group"` while the day cells use `role="gridcell"`. Per WAI-ARIA, a `gridcell` must be a descendant of a `row`, which must itself be inside a `grid`, so the current role chain is technically non-conformant.

In practice this works well with screen readers because navigation relies on the live region, but a spec-strict accessibility audit will flag the invalid hierarchy.

Two options:

- Promote the container to `role="grid"` and introduce proper `role="row"` wrappers for each week.
- Replace `role="gridcell"` on the day cells with a more appropriate role (for example `role="button"`), or remove the explicit role to match the `group` container.

This is separate from the P0 finding around the dead `aria-activedescendant` documentation .

chrisfalaska:
Caveat: vendor code still uses startOfDay from date-fns
  
The P0-5 fix consistently normalizes dates via .setHours(0, 0, 0, 0) in auro-calendar.js and auro-calendar-cell.js, but the vendored wc-range-datepicker still uses date-fns' startOfDay:

  - components/datepicker/src/vendor/wc-range-datepicker/range-datepicker-cell.js:4 — imports startOfDay
  - components/datepicker/src/vendor/wc-range-datepicker/range-datepicker-cell.js:69-70 — uses startOfDay
  - components/datepicker/src/vendor/wc-range-datepicker/range-datepicker-calendar.js:4 — imports startOfDay
  - components/datepicker/src/vendor/wc-range-datepicker/range-datepicker-calendar.js:49 — uses startOfDay

Outside the original bug's scope (which named only auro-calendar.js and auro-calendar-cell.js), and vendor code is typically not Auro-controlled — but flagging in case the DST-desync concern extends to vendor.

components/datepicker/src/auro-calendar.js
```diff
@@ -1064,20 +1174,18 @@ export class AuroCalendar extends RangeDatepicker {
...
+            const target = cells.find((cell) => cell.day && cell.day.date === targetDate);
+            if (target) {
+              this.setActiveCell(target.day.date);
+              this.focusActiveCell();
+            } else if (cells.length > 0) {
+              // Clamp to nearest focusable cell
+              const nearest = navDirection === 'next' ? cells[0] : cells[cells.length - 1];
+              this.setActiveCell(nearest.day.date);
+              this.focusActiveCell();
+            }
```
chrisfalaska: **What IS fixed**

`pickNearestCell(cells, targetTs, direction)` — `auro-calendar.js:517-543`

Linear scan picking the cell with minimum `Math.abs(cell.day.date - targetTs)`. Ties broken by navigation direction (`'next'` → later cell, `'prev'` → earlier). Returns `null` for empty cells.

Wired into the main grid keyboard handler `handleGridKeyDown`:

- ArrowLeft/Right path — `auro-calendar.js:1012-1027` → calls `pickNearestCell(cells, targetTs, navDir)` at line `1019`
- ArrowUp/Down path — `auro-calendar.js:1052-1066` → calls `pickNearestCell(cells, targetDate, navDirection)` at line `1059`

Tests added:

- `"should navigate ArrowDown across month boundary"` (test line ~5057)
- `"should navigate ArrowUp across month boundary"` (test line ~5084)

**What is NOT fixed**

The `handleMonthBoundary` handler (separate path that receives cross-month events bubbled up from child month components) still uses the broken pattern:

- `auro-calendar.js:1123` — `cells[cells.length - 1]` direct fallback
- `auro-calendar.js:1147` — `cells[0]` direct fallback
- `auro-calendar.js:1185-1188` — `navDirection === 'next' ? cells[0] : cells[cells.length - 1]`

```js
} else if (cells.length > 0) {
  const nearest = navDirection === 'next'
    ? cells[0]
    : cells[cells.length - 1];

  this.setActiveCell(nearest.day.date);
  this.focusActiveCell();
}
```

This is the exact off-by-one pattern the bug described — just in a different handler. When `handleMonthBoundary` fires (e.g., from a child month component's boundary event) and the target date isn't in the newly rendered cells, the user can still land on the extreme of the wrong month.

Tests at lines `6536-6578` cover this path but only assert `activeCellDate` is truthy — they would pass with either the buggy clamp or the correct nearest-cell pick, so they don't catch the regression.

**Recommendation**

Replace the three remaining `cells[0]` / `cells[cells.length - 1]` clamps in `handleMonthBoundary` (lines `1123`, `1147`, `1185-1188`) with calls to `this.pickNearestCell(cells, targetTs, navDirection)`, mirroring what `handleGridKeyDown` already does. Then strengthen the existing `handleMonthBoundary` tests to assert the resulting date, not just truthiness.

components/datepicker/test/auro-datepicker.test.js
chrisfalaska, line 8073: P1-8 — RESOLVED Tests: a11y depth is shallow

**Consensus: PARTIALLY FIXED ⚠️**

All three subagents converged on **PARTIALLY FIXED**. The resolution's premise is correct, but it only addresses 1 of the 3 original concerns.

### Concern (a) — `aria-activedescendant` pattern: GENUINELY FIXED ✅

- `grep` of `components/datepicker/src/` for `activedescendant`, `ActiveDescendant`, and `ariaActiveDescendantElement` returns zero hits.
- The "stale code removed, no test needed" claim is valid.

**Note:** The test file still has vestigial references (`auro-datepicker.test.js:841` comment, `:1278` test name), but they no longer assert anything about `aria-activedescendant` attributes—they test `tabindex` and `inert` instead. Worth a cleanup to avoid future confusion.

### Concern (b) — Live-region announcements tied to actions: PARTIALLY ADDRESSED ⚠️

**Now covered:**

- `auro-datepicker.test.js:8242-8268` — `"should populate live region when prev month button is clicked"` (asserts `liveRegion.textContent` includes month/year)
- `auro-datepicker.test.js:8271-8297` — same for next-month button
- `auro-datepicker.test.js:8905-8981` — `announceSelection` retry/race-condition tests

**Still missing:**

- No assertion that date selection (clicking a cell) triggers a live-region announcement with the picked date.
- No assertion that range-preview hover/focus announces the previewed range (despite this being P1-1's whole point).
- `announceFocusDebounced` mock at line `6722` only verifies the function is called, not the announcement content tied to a user action.

### Concern (c) — Focus restoration beyond `trigger.inert`: NOT ADDRESSED ❌

- `auro-datepicker.test.js:1239-1267` — `"should restore trigger inert and focus after fullscreen dialog closes"` only asserts:
  - `expect(dropdown.trigger.inert).to.be.false`
  - `expect(dropdown.isPopoverVisible).to.be.false`
- No assertion on `document.activeElement` or `shadowRoot.activeElement` confirming focus actually lands on the trigger element after close.

### Single axe-core scan still shallow

- `auro-datepicker.test.js:7896` — only one `expect(el).to.be.accessible()` call in the entire suite, with `ignoredRules: ['nested-interactive']`.
- No per-state axe scan (calendar open, range mid-selection, fullscreen mode, etc.).

### Recommendation for the PR

The "no need to add a11y tests" resolution overstates the fix. Two reasonable follow-ups:

1. Add a live-region content assertion after date selection (the most user-visible action, currently untested).
2. Strengthen the focus-restoration test to assert `document.activeElement === trigger` rather than just `trigger.inert === false`.
3. Optionally, add axe scans for the calendar-open and fullscreen states, not just the closed state.

components/datepicker/src/auro-calendar.js
chrisfalaska: Additionally, relating to **P2-7 — Calendar grid lacks `aria-roledescription="calendar"`:**

The original bug report said the calendar had `role="grid"` and just needed `aria-roledescription` added. But the actual element now has `role="group"`, not `role="grid"`.
That's a meaningful semantic difference:

- `role="grid"` implies a 2-D widget with arrow-key navigation and `gridcell` children — what calendars typically use
- `role="group"` is a generic container with no navigational semantics implied

If this was an intentional change (perhaps to avoid AT confusion with cell children that don't all conform to `role="gridcell"` requirements, or because the child months each have their own grid roles), it's fine — but it's a quieter change than "added `aria-roledescription`." If it was unintentional (or a regression from an earlier `role="grid"`), it's worth surfacing.

Worth a one-line PR comment: "Confirm `role="group"` (vs `role="grid"` mentioned in the original report) is intentional — grids have stronger AT semantics for 2D nav."

Otherwise the attribute itself is correctly added, correctly valued, correctly placed, and tested.

chrisfalaska (changes requested):
Independently verified each point raised in the adverserial review at https://github.com/AlaskaAirlines/auro-formkit/pull/1485#issuecomment-4793529237 against the fixes.

components/datepicker/src/auro-calendar.js
jason-capsule42: > The calendar container uses `role="group"` while the day cells use `role="gridcell"`. Per WAI-ARIA, a `gridcell` must be a descendant of a `row`, which must itself be inside a `grid`, so the current role chain is technically non-conformant.
> 
> In practice this works well with screen readers because navigation relies on the live region, but a spec-strict accessibility audit will flag the invalid hierarchy.
> 
> Two options:
> 
> * Promote the container to `role="grid"` and introduce proper `role="row"` wrappers for each week.
> * Replace `role="gridcell"` on the day cells with a more appropriate role (for example `role="button"`), or remove the explicit role to match the `group` container.
> 
> This is separate from the P0 finding around the dead `aria-activedescendant` documentation .

These extra attributes are not needed since the AT / screenreader are dependent on the aria live region instead of the native aria grid behavior.

jason-capsule42: > Additionally, relating to **P2-7 — Calendar grid lacks `aria-roledescription="calendar"`:**
> 
> The original bug report said the calendar had `role="grid"` and just needed `aria-roledescription` added. But the actual element now has `role="group"`, not `role="grid"`. That's a meaningful semantic difference:
> 
> * `role="grid"` implies a 2-D widget with arrow-key navigation and `gridcell` children — what calendars typically use
> * `role="group"` is a generic container with no navigational semantics implied
> 
> If this was an intentional change (perhaps to avoid AT confusion with cell children that don't all conform to `role="gridcell"` requirements, or because the child months each have their own grid roles), it's fine — but it's a quieter change than "added `aria-roledescription`." If it was unintentional (or a regression from an earlier `role="grid"`), it's worth surfacing.
> 
> Worth a one-line PR comment: "Confirm `role="group"` (vs `role="grid"` mentioned in the original report) is intentional — grids have stronger AT semantics for 2D nav."
> 
> Otherwise the attribute itself is correctly added, correctly valued, correctly placed, and tested.

This seems like a nit that doesn't seem worth it.

components/datepicker/src/auro-calendar.js
jason-capsule42: Code changes made to address this.

components/datepicker/test/auro-datepicker.test.js
jason-capsule42: Concern (a) — vestigial aria-activedescendant comments ✅ Cleaned up

[auro-datepicker.test.js:353-354](vscode-webview://03fttbe88tddn2dic8k2dne3e7kl0qdf5258kbfps8opbfgrc87p/components/datepicker/test/auro-datepicker.test.js#L353-L354) — comment rewritten to describe the actual behavior (host-level aria-disabled so AT browsing the gridcell sees the disabled state) without claiming an aria-activedescendant pattern that's no longer there.
[auro-datepicker.test.js:9480-9482](vscode-webview://03fttbe88tddn2dic8k2dne3e7kl0qdf5258kbfps8opbfgrc87p/components/datepicker/test/auro-datepicker.test.js#L9480-L9482) — test name and JSDoc retuned to "focus stays on grid wrapper" (the actual reason cells stay tabindex="-1"), replacing the inherited (aria-activedescendant pattern) parenthetical.
Concern (b) — date-selection live-region announcement ✅ Now covered

New test at [auro-datepicker.test.js:8341-8385](vscode-webview://03fttbe88tddn2dic8k2dne3e7kl0qdf5258kbfps8opbfgrc87p/components/datepicker/test/auro-datepicker.test.js#L8341-L8385): opens the picker, finds the cell for Jan 20, 2024 via a deterministic predicate, calls targetCell.handleTap(), waits past the announce double-rAF, and asserts the live region's textContent contains '20', 'January', '2024', and 'selected' (case-insensitive). Locks in the action→announcement contract.
The reviewer's note about range-preview hover/focus announcements is already exercised by the existing focus tests + the pickNearestCell paths — handleCellFocused → buildFocusAnnouncement → announceFocusDebounced is the same code path, just triggered by keyboard nav. The new date-click test pins the "action → live region content" guarantee at the most user-visible action point.
Concern (c) — focus restoration to trigger element ✅ Strengthened

[auro-datepicker.test.js:1239-1273](vscode-webview://03fttbe88tddn2dic8k2dne3e7kl0qdf5258kbfps8opbfgrc87p/components/datepicker/test/auro-datepicker.test.js#L1239-L1273): added expect(el.shadowRoot.activeElement).to.equal(input) after the existing trigger.inert === false / isPopoverVisible === false checks. Mirrors the same shadow-root active-element pattern used at lines 1855 and 2673 of the file. Also bumped the post-hideBib() settle to two rAFs since the focus-restoration is itself posted on requestAnimationFrame(...) inside the auroDropdown-toggled handler.
What I left in scope per the reviewer's "optionally" wording: per-state axe scans (open + fullscreen). The existing single axe scan covers the closed state; adding per-state scans across the bib-open lifecycle is a worthwhile follow-up but distinct enough that I'd recommend its own PR (the useAccessibleIt helper iteration is consistent across the suite and changing it at one point invites drift). Happy to add it on request.

jason-capsule42:
> Caveat: vendor code still uses startOfDay from date-fns
> 
> The P0-5 fix consistently normalizes dates via .setHours(0, 0, 0, 0) in auro-calendar.js and auro-calendar-cell.js, but the vendored wc-range-datepicker still uses date-fns' startOfDay:
> 
> * components/datepicker/src/vendor/wc-range-datepicker/range-datepicker-cell.js:4 — imports startOfDay
> * components/datepicker/src/vendor/wc-range-datepicker/range-datepicker-cell.js:69-70 — uses startOfDay
> * components/datepicker/src/vendor/wc-range-datepicker/range-datepicker-calendar.js:4 — imports startOfDay
> * components/datepicker/src/vendor/wc-range-datepicker/range-datepicker-calendar.js:49 — uses startOfDay
> 
> Outside the original bug's scope (which named only auro-calendar.js and auro-calendar-cell.js), and vendor code is typically not Auro-controlled — but flagging in case the DST-desync concern extends to vendor.

Would prefer to not refactor the Vendor code unless it's causing an actual issue. Is there one that can be reproduced?

chrisfalaska:
> > Caveat: vendor code still uses startOfDay from date-fns
> > The P0-5 fix consistently normalizes dates via .setHours(0, 0, 0, 0) in auro-calendar.js and auro-calendar-cell.js, but the vendored wc-range-datepicker still uses date-fns' startOfDay:
> > 
> > * components/datepicker/src/vendor/wc-range-datepicker/range-datepicker-cell.js:4 — imports startOfDay
> > * components/datepicker/src/vendor/wc-range-datepicker/range-datepicker-cell.js:69-70 — uses startOfDay
> > * components/datepicker/src/vendor/wc-range-datepicker/range-datepicker-calendar.js:4 — imports startOfDay
> > * components/datepicker/src/vendor/wc-range-datepicker/range-datepicker-calendar.js:49 — uses startOfDay
> > 
> > Outside the original bug's scope (which named only auro-calendar.js and auro-calendar-cell.js), and vendor code is typically not Auro-controlled — but flagging in case the DST-desync concern extends to vendor.
> 
> Would prefer to not refactor the Vendor code unless it's causing an actual issue. Is there one that can be reproduced?

Not that I saw, it was flagged during review so I wanted to raise it. Not a blocker.

chrisfalaska (approved):
Feedback addressed. Ship it! 🚀

PR #1512, opened 2026-06-25
Jbaker/input review updates

jason-capsule42:

## Review Checklist:

<details>
<summary>
RC Checklist
</summary>

### Browsers

Android

- [ ] Chrome
- [ ] Firefox

 iOS

- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Desktop

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### Scenarios

- [ ] Next React
- [ ] SvelteKit

</details><br>

-- Auro Design System Team

## Summary by Sourcery

Improve auro-input date formatting robustness, accessibility behavior, and localization examples.

New Features:
- Add new localization and ISO date format Storybook examples for auro-input.

Bug Fixes:
- Normalize date format tokens so uppercase format attributes correctly map to validation help text.
- Prevent cursor jumping in Safari for text-like inputs by restoring caret position after value updates.
- Ensure the password visibility toggle exposes its pressed state via aria-pressed for assistive technologies.
- Stabilize help-text DOM node identity across validity transitions so screen readers re-announce changes.
- Make ISO date formatting and date value formatting utilities return undefined for invalid or missing inputs instead of malformed strings or errors.

Enhancements:
- Consolidate and clarify BaseInput default property initialization for consistency and maintainability.

Documentation:
- Document default values for several auro-input properties and update date API examples to use ISO-formatted min/max values.

Tests:
- Add regression and edge-case tests covering help-text live region behavior, password toggle accessibility, format normalization, ISO/date utility error handling, and date formatting edge cases.


jason-capsule42:
@sourcery-ai review

components/input/src/base-input.js
```diff
@@ -1135,6 +1132,25 @@ export default class BaseInput extends AuroElement {
...
+    // formatter manages the cursor itself) doesn't get a competing write.
+    const { selectionStart } = this.inputElement;
+
+    if (this.setSelectionInputTypes.includes(this.type)) {
+      this.updateComplete.then(() => {
+        try {
+          this.inputElement.setSelectionRange(selectionStart, selectionStart);
+        } catch (error) { // eslint-disable-line
+          // Some input types (number/email in certain UAs) throw on
+          // setSelectionRange; swallow and let the native cursor stand.
```
rmenner, line 1150: What is the point of the try catch if the catch is commented out?

components/input/src/auro-input-util.js
```diff
@@ -21,6 +21,15 @@ function formatISODate(isoStr, format) {
   try {
     const date = dateFormatter.stringToDateInstance(isoStr);
 
+    // `stringToDateInstance` returns an `Invalid Date` for malformed strings
+    // and `null` for non-string input — it does NOT throw. Without this
+    // guard, `getFullYear()`/`getMonth()`/`getDate()` on an Invalid Date
+    // yield `NaN`, and `String(NaN).padStart(4, '0')` produces literals
+    // like "0NaN/0NaN/0NaN" instead of `undefined`.
+    if (!(date instanceof Date) || Number.isNaN(date.getTime())) {
+      return undefined;
```
rmenner, line 30: If this catches every error and returns undefined. A bug in the date code can look like normal bad input and never error out.

components/input/src/base-input.js
```diff
@@ -706,6 +706,15 @@ export default class BaseInput extends AuroElement {
     }
     this.inputId = this.id ? `${this.id}-input` : window.crypto.randomUUID();
 
+    // Normalize the format token to lowercase so case-mixed values supplied
+    // via attribute (e.g. `format="MM/DD/YYYY"`) hit the `dateFormatMap`
+    // lookup inside `setCustomHelpTextMessage`. Without this, an uppercase
+    // format silently misses the map and leaves `setCustomValidityForType`
+    // unset.
+    if (this.format) {
+      this.format = this.format.toLowerCase();
```
rmenner, line 715: Not sure it's a good idea to override the user provided format.I feel like we should make a new normalized format value to use inside the component. Also I don't think this will prevent someone from dynamically setting it later in the life cycle. It should probably be moved to willUpdate or changedProperties.

components/input/src/base-input.js
jason-capsule42: the comment calls out that if our cursor positioning fails the catch just let's the browser default behavior execute instead.

components/input/src/auro-input-util.js
jason-capsule42: The only error that could happen here is when someone tried to pass a string to the date attribute that was not a valid date. What is the practical concern here?

components/input/src/base-input.js
jason-capsule42: That would be a lot of code to refactor at this stage of the release prep. Open to this being logged as a new ticket for a future improvement.

components/input/src/auro-input-util.js
rmenner: It normalizes bad values to undefined. might be hard to catch errors

jason-capsule42: If we don't do that it will actually break the JS. What do you propose we do instead?

rmenner: We could narrow the check to just the documented cases (date === null for non-string input, isNaN(date.getTime()) for invalid date) so anything unexpected falls through instead of getting normalized. but fine to leave as is. I was just worried that it could cause issues later.

rmenner: Never mind. Didn't realize this was an opt-in helper function

rmenner (approved):

PR #1536, opened 2026-07-06
docs(menu, select): align keyboard & manual testing docs with implementation

chrisfalaska:
## Summary

Update for the `<auro-select>` and `<auro-menu>` keyboard-behavior tables and manual-testing checklists — bringing the published documentation into
alignment with what the components actually do today.

### `<auro-menu>` (`009c3cf85`)

- Replaces the stub "menu does not handle keyboard events" note with a real key/state table. `auro-menu` *does* own a `keydown` listener that handles `ArrowDown`,
  `ArrowUp`, `Enter`, and `Tab`; the parent (`auro-select` / `auro-combobox`) handles the rest (`Home`, `End`, `Escape`, type-ahead).
- Documents wrap-around, skip rules (disabled / `hidden` / `static` / `<hr>`), the empty-menu / all-non-interactive case, and the `loading`-state placeholder being
  inert.
- Clarifies `Enter` semantics — single-select re-selecting an already-selected option still fires `auroMenu-selectedOption` so the parent can react (e.g. close the
  dropdown); multi-select toggles.
- Bugfix: auro-menu's reset() previously left optionActive and the .active class stamped on the last-activated option, so a follow-up open/navigate reused a stale reference. reset() now clears both. Covered by a new unit test in auro-menu.test.js.

### `<auro-select>` (`28a08230f`)

- Renames "focused option" → "active option" throughout the key table to match the ARIA / implementation vocabulary (`aria-activedescendant`).
- Documents the fullscreen-bib case up front: focus moves to the dialog's close button, but keyboard nav still works via cross-shadow-root `aria-activedescendant`.
- Splits previously-conflated rows into explicit collapsed / expanded / multiSelect / fullscreen states for `Enter`, `Escape`, `Tab`, `Shift+Tab`, and `Space`.
- Documents the type-ahead buffer's default `typeaheadTimeoutMs` (500 ms) and that `Ctrl`/`Meta`/`Alt`-chorded keys are ignored so browser and OS shortcuts don't leak
  into the buffer.
- Documents the fullscreen `Escape` path: keydown bubbles from the close button to the select's keyboard strategy (same handler as popover); native `<dialog>`
  `cancel` → `auro-bib-cancel` is the fallback.
- Notes that `Tab` closes the bib and selects/toggles the `active` option only if one exists — no-op if nothing is active.
- Standardize "active option" terminology across docs and tests

### MANUAL_TESTING checklists

- `components/menu/test/MANUAL_TESTING.md` and `components/select/test/MANUAL_TESTING.md` reworked to match the new behavior tables — adds key-symbol legend, converts
  bare `[ ]` to markdown `- [ ]` checkboxes, adds rows for hidden/static skipping, hover-off vs. option-to-option active clearing, imperative-selected override
  behavior, and fullscreen focus target.

## Summary by Sourcery

Align auro-select and auro-menu keyboard behavior documentation and manual testing checklists with the components’ current keyboard, focus, and accessibility behavior.

Documentation:
- Expand and update auro-select keyboard-behavior docs to use active-option terminology, describe fullscreen dialog focus handling, Escape/Space/Tab semantics, and type-ahead behavior including modifier key handling.
- Replace the auro-menu keyboard behavior stub with a detailed key/state table describing Arrow, Enter, and Tab handling, including wrap-around, non-interactive item skipping, and loading-state behavior.

Tests:
- Significantly rework auro-select manual testing checklist to cover updated keyboard navigation, fullscreen behavior, type-ahead buffer details, property and method behaviors, edge cases, events, accessibility, and slots.
- Expand auro-menu and auro-menuoption manual testing checklists to validate keyboard interaction rules, selection behaviors, inherited and behavioral attributes, events, public APIs, nested menus, and accessibility.


components/select/docs/pages/customize.md
```diff
@@ -264,7 +264,7 @@
         <!-- AURO-GENERATED-CONTENT:END -->
         </auro-accordion>
         <auro-header level="3" id="disableOptions">Disable Option(s)</auro-header>
-        <p>The component may be rendered with one or more <code>disabled</code> options. When navigating the list of options with the keyboard or assistive technology to mark the next or previous option as active, disabled options will be skipped, jumping to the next enabled option.</p>
+        <p>The component may be rendered with one or more <code>disabled</code> options. When navigating the list of options with the keyboard or assistive technology to mark the next or previous option as active, disabled options will be skipped, jumping to the next active option.</p>
```
jason-capsule42, line 267: This is wrong. There are not multiple "active" options to jump to. There is only ever one active option and we jump to the next "enabled" option.

jason-capsule42: This pattern continues elsewhere in this commit. "active" was substituted for "enabled" in several places that are incorrect.

jason-capsule42 (changes requested):

components/select/docs/pages/customize.md
chrisfalaska: OK, this confused both me and the agent.

We use the word "active" for two completely different things in Menu and Select 

1. "The active option" -- the one menu item that's currently highlighted (`menu.optionActive`)
2. "Active options" (plural) -- the whole group of items the user can navigate to (everything not `disabled`, `hidden`, or `static`).

What's confusing is that the plural concept is called `isActive` an `getActiveOptions` in the code--the word "enabled" doesn't appear there at all. 

"Enabled" exists only in the docs as a concept the code names differently.

Ideally, we  rename the code so there's no confusion:

`isActive` becomes something like `isNavigable` & `getNavigableOptions`, and "active" is reserved exclusively for the single highlighted option (`menu.optionActive`).

chrisfalaska:
These changes will be committed in a different PR

PR #1541, opened 2026-07-07
docs(menu, select): API cleanup, missing examples, keyboard strategy

chrisfalaska:
## Summary

Focused pass on Menu and Select documentation, examples, and a handful of small feature and fix items surfaced along the way. No breaking changes.

### Menu

**Features**
- `feat(menu)`: add `event` attribute support for menu options, plus an example
- `feat(menu)`: declare `menuoption[persistent]` as a supported property (previously undocumented)

**Fixes**
- `fix(menu)`: `reset()` now clears active option state
- `fix(docs)`: clarify how disabled components behave in form submissions

**Docs & examples**
- New examples: reset, match-word, loading, dividers, static, no-match, event-attribute, persistent, and select-by-value
- Deprecate `menuoption[selected]` in favor of `menu[value]`
- Remove references to non-existent APIs, phantom keyboard support claims, and stale `allowDeselect` docs
- Correct size options and the `optionSelected` description
- Align public method docs with source markers
- Update keyboard interaction documentation

**Tests**
- Add coverage for direct `keydown` `ArrowUp`/`ArrowDown` navigation

### Select

- `fix(select)`: standardize `customColor` → `customcolor` attribute casing in menu options
- Update manual testing instructions for keyboard interactions


components/select/docs/partials/keyboard-behavior/keyEvents.md
```diff
@@ -189,7 +221,7 @@
...
       <td>
-        Extends the type-ahead buffer. When the buffer matches an enabled option's displayed text, the <code>focused</code> option advances to that option and the bib opens if it was collapsed. A keystroke that does not match any option leaves the bib state unchanged. Repeated keystrokes within <code>typeaheadTimeoutMs</code> extend the buffer; pressing the same character repeatedly cycles through matching options. See the <a href="customize.html#typeAhead">Type-Ahead</a> section on the Customize page.
+        Extends the type-ahead buffer. When the buffer matches an enabled option's displayed text, the <code>focused</code> option advances to that option and the bib opens if it was collapsed. A keystroke that does not match any option leaves the bib state unchanged. Repeated keystrokes within <code>typeaheadTimeoutMs</code> (default 500&nbsp;ms) extend the buffer; pressing the same character repeatedly cycles through matching options. Keys chorded with <code>Ctrl</code>, <code>Meta</code>, or <code>Alt</code> are ignored so browser and OS shortcuts do not leak into the buffer. See the <a href="customize.html#typeAhead">Type-Ahead</a> section on the Customize page.
```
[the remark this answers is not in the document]
chrisfalaska: This is a known issue, tracked here: https://github.com/AlaskaAirlines/auro-formkit/issues/1538

components/select/docs/partials/index/userStories.md
```diff
@@ -35,7 +35,7 @@
   <p>In some cases it is necessary to preset the value of the component as part of the initial render.</p>
   <auro-header level="4" id="us3a">Single Select</auro-header>
   <p>When a value is preset, the matching option in the menu will be marked as both <code>active</code> and <code>selected</code>.</p>
-  <p>If a preset value is defined that does not match any enabled option, the value is reset to <code>undefined</code>.</p>
+  <p>If a preset value is defined that does not match any option, the value is reset to <code>undefined</code>. (Value matching is not filtered by <code>disabled</code>, <code>hidden</code>, or <code>static</code> — a value that matches such an option is still applied.)</p>
```
jason-capsule42: Huh? Why are we allowing a disabled option to be preset? That seems like a bug?

components/select/docs/partials/keyboard-behavior/keyEvents.md
```diff
@@ -1,10 +1,14 @@
+<div class="note">
+  <p><strong>Note:</strong> The "Focused element" column below indicates the initial <code>event.target</code> — the element that has DOM focus when the key is pressed. The select's keyboard strategy listener is attached once to the <code>&lt;auro-select&gt;</code> host; it receives events via composed bubbling regardless of display mode. When the bib is displayed as a fullscreen dialog (see <code>fullscreenBreakpoint</code>), focus moves to the dialog's close button rather than remaining on the trigger, but keydown events still bubble across the shadow-root boundary up to the host listener, and <code>aria-activedescendant</code> continues to track the active option. An "active option" is one that is not <code>disabled</code>, <code>hidden</code>, or <code>static</code> — these are always skipped during navigation and type-ahead matching.</p>
```
jason-capsule42, line 2: Static options should absolutely not be skipped and defeats the entire purpose of their existence. This is a bug if that is how it actually works today.

components/select/docs/partials/keyboard-behavior/keyEvents.md
jason-capsule42, line 224: Was typeahead tested with the shift key and matching compares lowercased typing against lowercased option.textContent?

components/select/docs/partials/index/userStories.md
jason-capsule42: Do you mean that the value of the component is still set but the option is not selected in the menu? I think we need to talk through this behavior.

chrisfalaska: Agent's assessment:

What actually happens when a preset or programmatic value matches a `disabled`, `hidden`, or `static` option:

1. The value is retained (it is not reset to `undefined`), and
2. The option is marked as selected—it receives `selected`, `aria-selected="true"`, the checkmark, and its text is shown in the trigger.

Why: value matching in `auro-menu` is a pure equality check with no state filtering.

- Single-select:
  `const matchingOption = this.items.find((item) => item.value === this.value);`
- Multi-select:
  `this.items.filter((item) => valueArray.includes(item.value));`

Neither path consults `isOptionInteractive()`, which excludes `hidden`, `disabled`, and `static` options. That guard is only applied during user-driven interactions such as `makeSelection()`, `navigateOptions()`, and typeahead. The value is only reset to `undefined` when no matching option exists at all (`auroMenu-selectValueFailure`).

### Answering the reviewer's questions

**Q1: "Why are we allowing a disabled option to be preset? Is this a bug?"**

The current behavior is the result of an asymmetry. Users cannot click or keyboard-select `disabled`, `hidden`, or `static` options because those interaction paths are guarded. However, preset or programmatic values can resolve to those options because the value matching path does not apply the same guard.

Whether that is a bug is a design decision rather than an observable fact about the implementation.

**Q2: "Is the value set while the option remains unselected in the menu?"**

No. The matching option is genuinely marked as selected in the menu, including the checkmark and `aria-selected="true"`. The only exception is a `hidden` option, whose selected state is not visible because it is not rendered, even though the selection attributes are still applied.

### Coverage gap

There is currently no test covering a preset value that matches a `disabled`, `hidden`, or `static` option. Existing disabled-option tests exercise only the click and keyboard interaction paths, which are already guarded. Although the `disabledFixture` exists in the menu test fixtures, it is not currently referenced by any tests.

### Recommendation

The documentation accurately describes the current behavior. However, since multiple reviewers found it surprising, it is worth having a design discussion before merging.

The real question is not the wording but whether value matching should filter by option state to mirror user interaction. If so, that requires a code change (filtering non-interactive options during value matching, or resetting the value when the only match is non-interactive) along with new test coverage. If not, the documentation should explain the intent—for example, that preset values are trusted regardless of option state so externally provided values are not silently discarded.

components/select/docs/partials/keyboard-behavior/keyEvents.md
chrisfalaska: I think this may come down to another set of confusing definitions, similar to [this bug](https://github.com/AlaskaAirlines/auro-formkit/issues/1538). The code uses both `isActive` and `isOptionInteractive`.

Static options are non-interactive by design (headers/info rows) -- skipping is intentional, documented, and tested. If we want a selectable-but-persistent option, that's the `persistent` attribute.

The docs are behaviorally correct but terminologically inconsistent, which may be what is confusing?

- `keyEvents.md` calls it an "active option" ("not disabled, hidden, or static")
- `keyboard-behavior.md` calls the same concept an "interactive option"

Here is the agent's assessment:

> The disagreement comes down to what `static` means in `auro-menu`. The reviewer assumes static options should be navigable. In this component, `static` is defined as the opposite: a non-interactive option for section headers / informational rows. Being skipped is its entire purpose.
>
> **Evidence (consensus across all three agents)**
>
> **1. Arrow navigation skips static** — `components/menu/src/auro-menu-utils.js:79`
>
> ```js
> export function isOptionInteractive(option) {
>   return !option.hasAttribute('hidden') &&
>          !option.hasAttribute('disabled') &&
>          !option.hasAttribute('static');
> }
> ```
>
> **2. Type-ahead skips static** — `getActiveOptions()` (`selectUtils.js:27`) filters by the `isActive` getter (`auro-menuoption.js:154`):
>
> ```js
> get isActive() {
>   return !this.hasAttribute('hidden') &&
>          !this.disabled &&
>          !this.hasAttribute('static');
> }
> ```
>
> **3. This is documented & tested intent** — `menu/docs/.../customize.md`
>
> > "Add the `static` attribute to an `auro-menuoption` to make it non-interactive — the option renders but is skipped during keyboard navigation and cannot be selected. Useful for section headers or informational rows."
>
> Plus a passing test: `isOptionInteractive` should return `false` for static option.
>
> **The likely source of confusion**
>
> The reviewer may be conflating `static` with `persistent` — a separate attribute that keeps an option fully interactive/selectable while excluding it from search-filter text rewriting (e.g. an "Add new…" row). If someone wants an always-present, selectable option, `persistent` is the tool; `static` is intentionally inert.

components/select/docs/partials/keyboard-behavior/keyEvents.md
chrisfalaska: Yes on the matching, the typed key is lowercased (`auro-select.js:1093`) and compared against `option.textContent`, which is also lowercased in `_getOptionDisplayText` (`:1046-1048`), so it's lowercase-vs-lowercase throughout.

On Shift: it works as designed, the modifier guard only skips Ctrl/Meta/Alt, not Shift (`selectKeyboardStrategy.js:120`), so a Shift+letter arrives as an uppercase key and gets lowercased into the buffer.

That said, there was no test for this, so I've added one: https://github.com/AlaskaAirlines/auro-formkit/pull/1541/changes/dbe9bd877c485894cc36580cf9e36f0442465e1a

components/select/docs/partials/index/userStories.md
chrisfalaska: Confirmed: `dev` was not returning `undefined` for this case. The guard that produces that behavior was part of the menu service and got dropped during the reversion. Great catch.

The root cause is that when the menu can't match a programmatic value to a *selectable* option, it dispatches `auroMenu-selectValueFailure`. The combobox already listens for that event and clears `value`/`optionSelected`, but on `dev` the menu never rejected disabled options, so the failure path was never reached.

I first tried a direct port of `main`'s `isActive` check (`!hidden && !disabled && !static`), but that broke 14 combobox tests because this branch uses `hidden` as its type-ahead filtering mechanism. Hidden options are routinely hidden while still needing to remain valid programmatic selections.

The final fix introduces a separate `isSelectableByValue()` (`!disabled && !static`) used only for value matching. This preserves the existing combobox filtering behavior while rejecting disabled/static options, allowing them to fall through to the existing `auroMenu-selectValueFailure` path. That brings `dev` back in line with `main` for disabled/static values without changing the hidden-option behavior this branch depends on.

Fix was made in 3ea3d00be89b35b317378788007d5e77fd9216cd

components/select/docs/partials/keyboard-behavior/keyEvents.md
chrisfalaska: Discussed with @jason-capsule42 . We confirmed `static` is buggy on Prod and hides values when applied. WIll be making an [ADO ticket](https://dev.azure.com/itsals/Auro%20Design%20System/_workitems/edit/1593740) to track.

Also discussed renaming `static` to avoid confusion with the JavaScript reserved word.

Added the `static` attribute to the Menu documentation since it was missing public-facing. Change was made in 45700821db945657286600a7cd85b612be90e602

components/menu/src/auro-menu.js
```diff
@@ -399,11 +404,17 @@ export class AuroMenu extends AuroElement {
           // Defensive default: `formattedValue` can be undefined for unexpected value types,
           // and calling `.includes` on undefined would throw during reconciliation.
           const valueArray = this.formattedValue || [];
-          const matchingOptions = this.items ? this.items.filter((item) => valueArray.includes(item.value)) : [];
+          const matchingOptions = this.items ? this.items.filter((item) => isSelectableByValue(item) && valueArray.includes(item.value)) : [];
           newSelected = matchingOptions.length > 0 ? matchingOptions : undefined;
```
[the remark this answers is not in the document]
chrisfalaska: Fixed

components/select/docs/partials/index/userStories.md
chrisfalaska: A followup fix for multiSelect: https://github.com/AlaskaAirlines/auro-formkit/pull/1541/commits/33c920228ed7d392687f04b7e9808892c6eca943

components/menu/src/auro-menu.js
```diff
@@ -163,22 +165,26 @@ export class AuroMenu extends AuroElement {
 
       /**
        * Specifies the current active menuOption.
+       * @readonly
        */
       optionActive: {
         type: Object,
-        attribute: 'optionactive'
+        attribute: false
       },
```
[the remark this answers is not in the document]
chrisfalaska: Before the refactor, the property was declared in shorthand:

```js
optionActive: { type: Object },
```

That shorthand still creates an observed attribute—Lit lowercases the property name, so `optionActive` implicitly got an `optionactive` attribute from the very beginning. Nobody chose the attribute deliberately; it came free with the shorthand.

Who made it explicit: chrisfalaska, on Dec 19 2024, in commit `91dab986` — "refactor: updates menu to work with multiselect".

That refactor expanded all the shorthand property declarations into explicit objects. For `optionActive` they wrote out `attribute: 'optionactive'` (just spelling out the pre-existing default) and added an `@attr {object} optionactive` JSDoc line.

Why it's an inconsistency, not a deliberate API: In that same commit, the sibling `optionSelected` was changed from:

```js
{ type: Object }
```

to explicit `attribute: false`.

So the author consciously opted `optionSelected` out of attribute mapping—but left `optionActive` with the defaulted mapping. Both are internally-managed read-only element references with identical semantics, so there's no reason one would be settable-by-attribute and the other not.

Conclusion: The `optionactive` attribute was never an intentional, designed API—it's a leftover from Lit's default attribute derivation that got frozen into an explicit line (and documented) during the multiselect refactor, while its twin got the correct `attribute: false` treatment. That's exactly the inconsistency the Copilot reviewer flagged, which reinforces that removing it is the right call rather than a loss of intended functionality.

components/select/test/auro-select.test.js
```diff
@@ -3728,6 +3728,33 @@ function runTest(mobileView) {
...
+          await elementUpdated(el);
+
+          // Shift+B arrives as an uppercase 'B' in evt.key.
+          el.dispatchEvent(new KeyboardEvent('keydown', { key: 'B', shiftKey: true }));
+          await elementUpdated(el);
+
+          expect(el.typeaheadBuffer).to.equal('b');
+          await expect(el.menu.optionActive).to.exist;
+          await expect(el.menu.optionActive.value).to.equal('banana');
+          await expect(el.menu.optionActive.textContent.trim()).to.equal('Banana');
```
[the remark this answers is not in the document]
chrisfalaska: The await is inert. Chai's `.to.exist` (a getter) and `.to.equal()` (a method) are fully synchronous. The assertion fires and throws before `await` ever sees its result. Awaiting the returned non-thenable `Assertion` object just burns one microtask tick—zero effect on pass/fail, timing, or correctness.

2. `expect(x).to.exist` (no parens) is **not** a bug. `.exist` is intentionally a Chai getter that asserts on property access. This is correct idiom, and the `await` doesn't interact with it.

3. Copilot's "hides mistakes by implying async behavior" claim is unsound. The `await` cannot mask a failed assertion—Chai throws synchronously on access, before the `await` resolves.

4. Copilot's "inconsistent with surrounding tests" claim is backwards. `await expect(...)` is the dominant convention in this file—397 uses vs. 119 plain `expect(...)`, including many identical `await expect(x).to.exist` lines elsewhere (e.g. 62, 1209, 2087). Removing the `await` would make these three lines less consistent with the file, not more.

Bottom line: The fix is safe either way (it's a cosmetic nit), but both of Copilot's justifications are factually wrong. If you value spec-purity, drop the `await`; if you value matching this file's prevailing style, keep it—which is the stronger argument here. Not worth blocking on, and I'd reject the suggestion as stated.

components/menu/docs/pages/css-only.md
```diff
@@ -68,45 +68,49 @@
...
 
 <auro-header level="3">Nested submenu support</auro-header>
 
-<p><code>auro-menu</code> detects nested <code>auro-menu</code> elements, applies the correct <code>role="group"</code>, computes indentation per level, and propagates the shared menu service down the tree. With plain HTML, you must apply nested ARIA roles and indentation manually, and there is no built-in coordination of focus or selection between parent and child lists.</p>
+<p><code>auro-menu</code> detects nested <code>auro-menu</code> elements, applies the correct <code>role="group"</code>, and computes indentation per level. Each nested menu maintains its own selection state — the root menu's <code>value</code> and <code>optionSelected</code> are not updated when an option inside a nested menu is selected. With plain HTML, you must apply nested ARIA roles and indentation manually.</p>
```
jason-capsule42: This confuses me - the root menu should always know which option is selected even if nested. Is this not how it works any more?

components/menu/docs/pages/customize.md
```diff
@@ -9,15 +9,20 @@
       <auro-anchorlink fluid href="#restrictedWidth" class="level2 body-xs">Restricted Width</auro-anchorlink>
       <auro-anchorlink fluid href="#scroll" class="level2 body-xs">Scroll</auro-anchorlink>
       <auro-anchorlink fluid href="#nocheckmark" class="level2 body-xs">No Checkmark</auro-anchorlink>
+      <auro-anchorlink fluid href="#dividers" class="level2 body-xs">Dividers</auro-anchorlink>
+      <auro-anchorlink fluid href="#static" class="level2 body-xs">Static Options</auro-anchorlink>
+      <auro-anchorlink fluid href="#matchWord" class="level2 body-xs">Match Word</auro-anchorlink>
+      <auro-anchorlink fluid href="#persistent" class="level2 body-xs">Persistent Option</auro-anchorlink>
+      <auro-anchorlink fluid href="#nomatch" class="level2 body-xs">No Match Placeholder</auro-anchorlink>
```
jason-capsule42, line 18: These should be under "Behavior" since they do affect how the overall menu functions. Where as line 12 only affects appearance with no functional change.

components/menu/docs/pages/getting-started.md
```diff
@@ -22,10 +22,13 @@
       <auro-anchorlink fluid href="#options" class="level2 body-xs">options</auro-anchorlink>
       <auro-anchorlink fluid href="#index" class="level2 body-xs">index</auro-anchorlink>
       <auro-anchorlink fluid href="#functions">Functions</auro-anchorlink>
-      <auro-anchorlink fluid href="#makeSelection" class="level2 body-xs">makeSelection()</auro-anchorlink>
```
jason-capsule42, line 25: Was this function actually removed or just deprecated?

components/menu/docs/pages/voiceover.md
```diff
@@ -24,7 +24,7 @@
 <div class="note"><strong>Note:</strong> <code>&lt;auro-menuoption&gt;</code> is not directly focusable — <code>tabindex="-1"</code> keeps it out of the tab order. In a select or combobox context, focus stays on the parent trigger and the active option is pointed at via <code>aria-activedescendant</code> (using <code>ariaActiveDescendantElement</code> to cross the shadow-root boundary).</div>
 
 <auro-header level="3" id="voiceOverSelection">Selection</auro-header>
-<p>When an option is selected via Enter or Space:</p>
+<p>When an option is selected via Enter (or Tab, which selects and moves focus out of the menu):</p>
```
jason-capsule42: I wonder if we should even document keyboard behavior since no key events are ever actually occuring on any part of the menu. Thoughts?

components/menu/docs/pages/voiceover.md
```diff
@@ -34,8 +34,8 @@
 <ol>
   <li><strong>Enter the menu:</strong> Focus moves into the listbox; VoiceOver announces role and item count</li>
   <li><strong>Navigate options:</strong> Use Arrow Down/Up keys; VoiceOver announces each option's label, role, and position</li>
-  <li><strong>Select an option:</strong> Press Enter or Space; VoiceOver confirms the selection</li>
-  <li><strong>Exit:</strong> Tab out of the menu or press Escape (when used within a dropdown context)</li>
+  <li><strong>Select an option:</strong> Press Enter; VoiceOver confirms the selection</li>
+  <li><strong>Exit:</strong> Tab out of the menu (Tab also selects the active option). When the menu is used inside a dropdown context, the parent handles Escape.</li>
```
jason-capsule42: Similar comment about key events here.

components/menu/docs/api.md
```diff
@@ -66,7 +72,8 @@ The `auro-menuoption` element provides users a way to define a menu option.
 | `layout`      |               |           | `string`  |         |                                                  |
 | `noCheckmark` | `noCheckmark` |           | `boolean` | false   |                                                  |
 | `noMatch`     | `nomatch`     |           | `boolean` | false   | When true, marks this option as the "no matching results" placeholder shown by combobox<br />when the user's input does not match any available options. |
-| `selected`    | `selected`    |           | `boolean` | false   |                                                  |
+| `persistent`  | `persistent`  |           | `boolean` | false   | When true, this option is excluded from `matchWord` DOM rewriting — useful for utility rows (e.g., "Add new…") that must render identically regardless of the current filter. |
+| `selected`    | `selected`    |           | `boolean` | false   | **Deprecated.** Use the `value` attribute on `auro-menu` to set the selected option when the menu renders (or call `menu.selectByValue(value)` programmatically). Support for the child-level `selected` attribute will be removed in a future major release. |
```
jason-capsule42, line 76: What is the reason we would want to remove `selected` off of the actual menuoption? Isn't this critical for VoiceOver screenreader experience?

components/menu/src/auro-menu-utils.js
```diff
@@ -43,6 +43,18 @@ export function arrayConverter(value) {
...
 
+/**
+ * Serializes a multi-select value array back into the String `value` property.
+ * An empty (or missing) array collapses to `undefined` so an emptied selection
+ * clears `value` rather than reflecting a `"[]"` attribute.
+ * @private
+ * @param {Array<string>|undefined} values - The selected values.
+ * @returns {string|undefined} JSON string of the values, or undefined when empty.
+ */
+export function serializeMultiSelectValue(values) {
```
jason-capsule42, line 54: What is the issue that exists in Dev today that necessitates this late code change?

components/menu/src/auro-menu-utils.js
```diff
@@ -82,6 +94,20 @@ export function isOptionInteractive(option) {
...
+/**
+ * Validates if an option may be selected by matching a programmatic value.
+ * Unlike `isOptionInteractive`, `hidden` is allowed: the combobox toggles
+ * `hidden` as its type-ahead filter, so a filtered-out option is still a
+ * valid programmatic selection. Only disabled and static options — which are
+ * never selectable — are rejected.
+ * @param {HTMLElement} option - The option to check.
+ * @returns {boolean} True if option can be selected by value.
+ */
+export function isSelectableByValue(option) {
```
jason-capsule42, line 106: What is the issue that exists in Dev today that necessitates this late code change?

components/menu/src/auro-menuoption.js
```diff
@@ -56,6 +57,7 @@ export class AuroMenuOption extends AuroElement {
     this.noCheckmark = false;
     this.disabled = false;
     this.noMatch = false;
+    this.persistent = false;
```
jason-capsule42, line 60: It's fine to leave this for now but doing any of these causes an unnecessary Lit lifecycle churn.

Any boolean property that does not initially get set to a `false` value is already false. Setting it here makes Lit see it as setting the value again triggering a new update.

components/menu/src/auro-menuoption.js
```diff
@@ -77,6 +79,12 @@ export class AuroMenuOption extends AuroElement {
         type: Boolean,
         reflect: true
       },
+
+      /**
+       * **Deprecated.** Use the `value` attribute on `auro-menu` to set the selected option when the menu renders (or call `menu.selectByValue(value)` programmatically). Support for the child-level `selected` attribute will be removed in a future major release.
+       *
+       * @deprecated Use the `value` attribute on `auro-menu` instead.
```
jason-capsule42, line 86: Same question - why would we remove this? Isn't it critical for the screenreader announcement in VoiceOver?

components/menu/docs/pages/css-only.md
chrisfalaska: The doc was wrong. It should say the root menu does track selection across nested menus:

"auro-menu detects nested `auro-menu` elements, applies `role="group"`, and computes indentation per level. Selecting an option in a nested menu updates the root menu's `value` and `optionSelected`, since the root treats all descendant options as one flat list."

I also added an additional test to assert "should update the root menu value/optionSelected when a nested option is clicked."

components/menu/docs/pages/customize.md
chrisfalaska: Fixed

components/menu/docs/pages/getting-started.md
chrisfalaska: It wasn't removed or deprecated. `makeSelection()` was marked `@private` in `auro-menu.js` (previously `@protected`) as part of the menu reversion, and is still used internally (keyboard + click handlers). It was mistakenly documented as a public method; this just removes it from the public docs to match its `@private` status. The public selection API is `selectByValue()` / `reset()`.

Please let me know if that assessment is incorrect.

components/menu/src/auro-menu-utils.js
chrisfalaska: It was an effect of switching back to the "undefined" check in disabled and static. A bug arose in multiselect and the fix the agent generated duplicated code in 3 places so i wrote a util to DRY it. I ran it against all of the tests, no errors.

components/menu/src/auro-menu-utils.js
chrisfalaska: The switch back to "undefined" for static and disbaled

components/menu/docs/pages/voiceover.md
chrisfalaska: Yea, interesting idea. The menu does register a real keydown listener on its host and handles ArrowUp/Down, Enter, Tab (auro-menu.js:335 → handleKeyDown at auro-menu.js:879). So "no key events ever" is incorrect.

But of course the menu is not focusable by default.

Instead, I updated the doc to describe the parent-managed model that actually applies under VoiceOver.

components/menu/docs/pages/voiceover.md
chrisfalaska: Yea, interesting idea. The menu does register a real keydown listener on its host and handles ArrowUp/Down, Enter, Tab (auro-menu.js:335 → handleKeyDown at auro-menu.js:879). So "no key events ever" is incorrect.

But of course the menu is not focusable by default.

Instead, I updated the doc to describe the parent-managed model that actually applies under VoiceOver.

components/menu/docs/api.md
chrisfalaska: It's there it's just deprecated, so it still works:

`Use the `value` attribute on `auro-menu` to set the selected option when the menu renders`

Screen readers announce selection via the `aria-selected` ARIA attribute, not the HTML selected attribute -- and `aria-selected` is always owned and set by the parent `auro-menu`, independent of how you preselect.

> **Agent's assessment:**
>
> * Screen readers read `aria-selected`, not `selected`. `voiceover.md` and `accessibility.md` both state the "selected" announcement fires off `aria-selected="true"`. The HTML `selected` attribute is never consumed by AT.
>
> * The parent menu sets both attributes together, in the same cycle. In `auro-menu.js:561-570`, whenever `value`/`optionSelected` changes, the menu sets `aria-selected` and adds/removes `selected` on each option. So `menu.value` / `selectByValue()` produce identical ARIA state to the old child attribute.
>
> * On the option side, `selected` is purely reactive. `auro-menuoption.js:202-203` syncs `aria-selected` from the `selected` property, and `setSelected()` (`:219`) is a pure setter the menu calls. The option can't own or diverge its own state—by design, per the distributed-state revert (`9be1e1dd4`).
>
> Why deprecate it: single source of truth. Post-5.9, the menu owns all selection state. The child `selected` attribute created a second, redundant write path that could desync from `auro-menu.optionSelected` (the JSDoc on `handleClick` calls out the exact single-select re-click desync it prevents). Consolidating on `menu.value` / `selectByValue()` removes that dual-state footgun while keeping ARIA output byte-for-byte identical.
>
> So: no a11y regression—migrating from `<auro-menuoption selected>` to `<auro-menu value="…">` yields the same `aria-selected` and the same VoiceOver announcement.

components/menu/src/auro-menuoption.js
jason-capsule42: Answered earlier in another comment

components/menu/src/auro-menuoption.js
chrisfalaska: I ran a test and the Constructor init does add these keys to `changedProperties` on the first `updated()` (measured: `["selected","disabled"]` vs. `[]`).

That's a semantic side effect, however. The `updated()` guards fire once on first render, but not "churn." No extra renders, just a populated first-render change set .

The agent *did* recognize a potentially latent bug, however:

`updated()` at `auro-menuoption.js:203` calls `this.selected.toString()` unguarded. If you remove `this.selected = false`, `selected` becomes `undefined` and that line is a `TypeError` waiting to happen.

The other four (`noCheckmark`, `disabled`, `noMatch`, `persistent`) are safe to remove--verified that all consumers (`auro-menu`, `combobox`, `select`) use `hasAttribute()` / truthy checks, with no `=== false` / `!== false` / `typeof` comparisons anywhere.

components/menu/src/auro-menuoption.js
chrisfalaska: It's there it's just deprecated, so it still works:

`Use the `value` attribute on `auro-menu` to set the selected option when the menu renders`

Screen readers announce selection via the `aria-selected` ARIA attribute, not the HTML selected attribute -- and `aria-selected` is always owned and set by the parent `auro-menu`, independent of how you preselect.

> **Agent's assessment:**
>
> * Screen readers read `aria-selected`, not `selected`. `voiceover.md` and `accessibility.md` both state the "selected" announcement fires off `aria-selected="true"`. The HTML `selected` attribute is never consumed by AT.
>
> * The parent menu sets both attributes together, in the same cycle. In `auro-menu.js:561-570`, whenever `value`/`optionSelected` changes, the menu sets `aria-selected` and adds/removes `selected` on each option. So `menu.value` / `selectByValue()` produce identical ARIA state to the old child attribute.
>
> * On the option side, `selected` is purely reactive. `auro-menuoption.js:202-203` syncs `aria-selected` from the `selected` property, and `setSelected()` (`:219`) is a pure setter the menu calls. The option can't own or diverge its own state—by design, per the distributed-state revert (`9be1e1dd4`).
>
> Why deprecate it: single source of truth. Post-5.9, the menu owns all selection state. The child `selected` attribute created a second, redundant write path that could desync from `auro-menu.optionSelected` (the JSDoc on `handleClick` calls out the exact single-select re-click desync it prevents). Consolidating on `menu.value` / `selectByValue()` removes that dual-state footgun while keeping ARIA output byte-for-byte identical.
>
> So: no a11y regression—migrating from `<auro-menuoption selected>` to `<auro-menu value="…">` yields the same `aria-selected` and the same VoiceOver announcement.

components/menu/src/auro-menuoption.js
jason-capsule42: Sounds like future improvements to me.

jason-capsule42 (approved):

PR #1554, opened 2026-07-15
fix(combobox): address displayValue synthesis and focus handling issues

jason-capsule42:

## Review Checklist:

<details>
<summary>
RC Checklist
</summary>

### Browsers

Android

- [ ] Chrome
- [ ] Firefox

 iOS

- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Desktop

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### Scenarios

- [ ] Next React
- [ ] SvelteKit

</details><br>

-- Auro Design System Team

## Summary by Sourcery

Fix combobox display value rendering and focus behavior for programmatic value updates and framework integrations.

Bug Fixes:
- Ensure combobox display value is rendered when a value is set before options are loaded by synthesizing a temporary displayValue from the current value.
- Prevent combobox focus handler from redirecting focus when focus originates from shadow-DOM children, avoiding focus loops with frameworks.
- Clear stale programmatic filter flags on real user input so the bib opens correctly on the first typed character after mount.
- Guard auro-input displayValue slot inspection against early lifecycle timing so hasDisplayValueContent is correctly set even before the shadow root slot exists, avoiding hidden display content.

Enhancements:
- Always re-evaluate auro-input displayValue slot state after updating trigger text so the visual wrapper reflects current content.
- Add tests covering programmatic value sync behavior, clear button focus behavior without combobox focus, and focus-in handling from shadow children.

Documentation:
- Add a detailed post-mortem documenting the displayValue and focus regressions, root causes, and implemented fixes.

### DIFF
--- a/components/combobox/src/auro-combobox.js
+++ b/components/combobox/src/auro-combobox.js
@@ -732,7 +732,6 @@ export class AuroCombobox extends AuroElement {
       // hidden when input.value is set in the same tick as the append.
       if (typeof this.input.checkDisplayValueSlotChange === 'function') {
         this.input.checkDisplayValueSlotChange();
-        this.input.requestUpdate();
       }
     } else if (this.value) {
       // No optionSelected yet (e.g. setMenuValue was called but the menu's
--- a/components/combobox/src/auro-combobox.js
+++ b/components/combobox/src/auro-combobox.js
@@ -739,6 +739,14 @@ export class AuroCombobox extends AuroElement {
       // auroMenu-selectedOption event hasn't fired yet, or we're on remount
       // with no options loaded). Synthesize a displayValue from the value so
       // the overlay isn't blank.
+      //
+      // NOTE: This renders the raw machine value (e.g. "SEA"), not the
+      // human-readable label (e.g. "Seattle, WA"). For consumers where
+      // value ≠ label, this produces a brief flash of the raw value until
+      // the option elements load and updateTriggerTextDisplay re-runs with
+      // the correct label. This is intentionally preferred over a blank
+      // state — the displayValueInTrigger.remove() cleanup above ensures
+      // this placeholder is replaced once the real option is available.
       const syntheticDV = document.createElement('span');
       syntheticDV.setAttribute('slot', 'displayValue');
       syntheticDV.textContent = this.value;
@@ -917,6 +925,10 @@ export class AuroCombobox extends AuroElement {
     // Prevent dropdown from closing on focus loss since menu content is slotted
     // from combobox's light DOM and won't be detected by dropdown's contains() check.
     this.dropdown.noHideOnThisFocusLoss = true;
+
+    // Suppress the dropdown's generic focus restoration on close — the combobox
+    // manages its own focus via setClearBtnFocus / setInputFocus / keyboard strategy.
+    this.dropdown.noFocusRestoreOnClose = true;
 
     // Listen for the ID to be added to the dropdown so we can capture it and use it for accessibility.
     this.dropdown.addEventListener('auroDropdown-idAdded', (event) => {
--- a/components/combobox/src/auro-combobox.js
+++ b/components/combobox/src/auro-combobox.js
@@ -1276,6 +1276,9 @@ export class AuroCombobox extends AuroElement {
       // only — fresh user selections take the existing hideBib path.
       if (isEcho && this.menu.optionSelected) {
         this._programmaticFilterRefresh = true;
+        this.updateComplete.then(() => {
+          this._programmaticFilterRefresh = false;
+        });
       }
     });
 
@@ -1506,15 +1509,6 @@ export class AuroCombobox extends AuroElement {
    * @returns {void}
    */
   handleTriggerInputValueChange(event) {
-    // Clear a stale _programmaticFilterRefresh when the user is genuinely
-    // interacting. On mount, updated('value') sets the flag, but the early
-    // return at `this.value === this.input.value` exits before clearing it.
-    // Without this, the first keystroke after mount fails to set _userTyped
-    // and the bib doesn't open.
-    if (this._programmaticFilterRefresh && this.componentHasFocus) {
-      this._programmaticFilterRefresh = false;
-    }
-
     // 'input' fires for every user-initiated value change — typing, paste,
     // IME composition end, dead-key composition, drag-drop. Flip _userTyped
     // here so updated('availableOptions') auto-opens the bib for sources
@@ -1536,10 +1530,6 @@ export class AuroCombobox extends AuroElement {
 
     if (this.dropdown.isBibFullscreen && this.input.value && this.input.value.length > 0 && this.dropdown.isPopoverVisible) {
       this.setInputFocus();
-    }
-
-    if (this._programmaticFilterRefresh) {
-      this._programmaticFilterRefresh = false;
     }
   }
 
@@ -1739,6 +1729,12 @@ export class AuroCombobox extends AuroElement {
       // setting the flag unconditionally here masks the user-typed open path.
       if (!this._userTyped) {
         this._programmaticFilterRefresh = true;
+        // Self-clear after this update cycle completes. This collapses
+        // the previous scattered clear-points into one and prevents the
+        // flag from surviving into the next user interaction.
+        this.updateComplete.then(() => {
+          this._programmaticFilterRefresh = false;
+        });
       }
 
       if (this.input.value !== this.value) {
@@ -1827,10 +1823,6 @@ export class AuroCombobox extends AuroElement {
         }
       } else if (!this.dropdown.isBibFullscreen) {
         this.hideBib();
-      }
-
-      if (this._programmaticFilterRefresh) {
-        this._programmaticFilterRefresh = false;
       }
     }
 
--- a/components/combobox/src/auro-combobox.js
+++ b/components/combobox/src/auro-combobox.js
@@ -752,7 +752,6 @@ export class AuroCombobox extends AuroElement {
       this.input.appendChild(syntheticDV);
       if (typeof this.input.checkDisplayValueSlotChange === 'function') {
         this.input.checkDisplayValueSlotChange();
-        this.input.requestUpdate();
       }
     }
 
--- a/components/combobox/src/auro-combobox.js
+++ b/components/combobox/src/auro-combobox.js
@@ -1546,7 +1546,7 @@ export class AuroCombobox extends AuroElement {
     this.addEventListener('focusin', (event) => {
       this.touched = true;
 
-      if (event.target === this) {
+      if (event.composedPath()[0] === this) {
         this.focus();
       }
     });
--- a/components/combobox/src/auro-combobox.js
+++ b/components/combobox/src/auro-combobox.js
@@ -724,14 +724,28 @@ export class AuroCombobox extends AuroElement {
       const displayValueEl = this.menu.optionSelected.querySelector("[slot='displayValue']");
       if (displayValueEl) {
         this.input.appendChild(displayValueEl.cloneNode(true));
-        // auro-input's hasDisplayValueContent is non-reactive; nudge it to
-        // re-evaluate the slot and re-render so the displayValue wrapper
-        // gets its hasContent class. Without this, the apple/icon stays
-        // hidden when input.value is set in the same tick as the append.
-        if (typeof this.input.checkDisplayValueSlotChange === 'function') {
-          this.input.checkDisplayValueSlotChange();
-          this.input.requestUpdate();
-        }
+      }
+
+      // auro-input's hasDisplayValueContent is non-reactive; nudge it to
+      // re-evaluate the slot and re-render so the displayValue wrapper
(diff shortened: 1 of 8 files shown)


components/combobox/src/auro-combobox.js
```diff
@@ -724,14 +724,28 @@ export class AuroCombobox extends AuroElement {
...
+        this.input.requestUpdate();
+      }
+    } else if (this.value) {
+      // No optionSelected yet (e.g. setMenuValue was called but the menu's
+      // auroMenu-selectedOption event hasn't fired yet, or we're on remount
+      // with no options loaded). Synthesize a displayValue from the value so
+      // the overlay isn't blank.
+      const syntheticDV = document.createElement('span');
+      syntheticDV.setAttribute('slot', 'displayValue');
+      syntheticDV.textContent = this.value;
```
jordanjones243, line 751: This renders the *machine value*, not a label. For flight-search (airport code == what we display) that's fine, but this path runs for **every** combobox consumer — where `value ≠ label` (e.g. value=`"SEA"`, label=`"Seattle, WA"`) users will see the raw value flash before options load, then swap. Still better than the previous blank state, but could we (a) add a comment noting this span is a placeholder showing the raw value until options load, and (b) confirm the flash is acceptable for non-flight-search consumers? (The `displayValueInTrigger.remove()` teardown above correctly prevents accumulation 👍.)

components/combobox/src/auro-combobox.js
```diff
@@ -1481,6 +1495,15 @@ export class AuroCombobox extends AuroElement {
    * @returns {void}
    */
   handleTriggerInputValueChange(event) {
+    // Clear a stale _programmaticFilterRefresh when the user is genuinely
+    // interacting. On mount, updated('value') sets the flag, but the early
+    // return at `this.value === this.input.value` exits before clearing it.
+    // Without this, the first keystroke after mount fails to set _userTyped
+    // and the bib doesn't open.
+    if (this._programmaticFilterRefresh && this.componentHasFocus) {
```
jordanjones243: This works, but it's the exact pattern Lesson #1 calls out — a leak-prone one-shot flag. We now have three clear-points + one set-point. Rather than keep adding clear-on-interaction guards, could we make the flag **self-clearing** on the update cycle it's meant to guard (e.g. schedule the clear via `this.updateComplete.then(...)` right after `updated('value')` sets it)? That collapses the clear-points into one and removes the "flag survives into the next interaction" bug class. Fine as a fast-follow so it doesn't block the flight-search fix.

Separately: this gate uses `componentHasFocus`, which Lesson #3 notes is unreliable during mobile fullscreen transitions. Probably fine since focus is settled by keystroke time — but that's why I'd like the Bug #4 test to run in fullscreen mode too (see summary).

components/combobox/src/auro-combobox.js
```diff
@@ -1520,10 +1543,12 @@ export class AuroCombobox extends AuroElement {
       inputResolver: (comp, ctx) => (ctx.isModal && comp.inputInBib ? comp.inputInBib : comp.input),
     });
 
-    this.addEventListener('focusin', () => {
+    this.addEventListener('focusin', (event) => {
       this.touched = true;
 
-      this.focus();
+      if (event.composedPath()[0] === this) {
```
jordanjones243, line 1560: Nice — this is the cleanest fix in the set. One low-risk note: focus no longer delegates to the input for slotted *clickable* children (composedPath[0] is that node, not the host), whereas the old unconditional call would have. Just flagging the narrowing; I don't think it needs a change.

components/input/src/auro-input.js
```diff
@@ -328,7 +328,16 @@ export class AuroInput extends BaseInput {
...
     // counts as content. The prior nodes[0].tagName === 'SLOT' recursion
     // discarded any siblings past the forwarder.
-    const slot = this.shadowRoot.querySelector('slot[name="displayValue"]');
+    const slot = this.shadowRoot?.querySelector('slot[name="displayValue"]');
+    if (!slot) {
+      // Shadow root not yet rendered — fall back to checking light-DOM
+      // children so hasDisplayValueContent isn't stuck false after an
+      // early call from auro-combobox during mount.
+      this.hasDisplayValueContent = Boolean(
+        this.querySelector('[slot="displayValue"]:not(slot)')
```
jordanjones243: Minor: the fallback matches an *element* with `slot="displayValue"`, while the primary path counts text nodes via `assignedNodes({flatten:true})`. Irrelevant for the synthesized `<span>`, but a latent gap if anyone ever slots bare text into displayValue.

jordanjones243 (changes requested):
Nice work — the diagnosis here is solid and the post-mortem made this easy to review. Each of the four fixes maps to a real root cause, and I appreciate that they're additive guards rather than behavior removal. I'm comfortable with the direction; my line-level notes are inline. Two things belong up here because they aren't tied to a single line:

**✅ What's great**
- The `focusin` composedPath guard is the standout — since the host has no `tabindex`/`delegatesFocus`, `composedPath()[0] === this` fires exactly in the framework-drives-focus-to-host case where we *want* delegation, and correctly skips when a shadow child is the origin. Clean and well-targeted.
- The auro-input null-safe slot query with a real default (light-DOM fallback) instead of a bare `return` is exactly right per your own Lesson #5.

**🔶 Tests don't cover the two headline bugs**
The three new tests all assert *negative* programmatic behavior (`_userTyped` stays false, clear button not focused, host `focus()` not called), which is good — but the two user-facing regressions aren't verified:
- **Bug #1 (blank overlay → populated):** no test asserts the displayValue overlay actually renders when `value` is set before options load.
- **Bug #4 (bib opens on first keystroke after mount):** the existing test checks `_userTyped` stays *false* on a programmatic sync; nothing verifies it flips *true* on the first real keystroke — which is the actual bug. Ideally exercised in **mobile fullscreen** too, given the `componentHasFocus` caveat noted inline.

components/combobox/src/auro-combobox.js
jason-capsule42: Addressed in an earlier commit.

components/input/src/auro-input.js
jason-capsule42: Addressed in an earlier commit.

components/input/src/auro-input.js
```diff
@@ -328,7 +328,16 @@ export class AuroInput extends BaseInput {
...
+      );
+      return;
+    }
```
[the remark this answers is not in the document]
jason-capsule42: Addressed in an earlier commit.

components/input/src/auro-input.js
```diff
@@ -323,15 +323,44 @@ export class AuroInput extends BaseInput {
...
+      // children so hasDisplayValueContent isn't stuck false when called
+      // before firstUpdated (e.g. from auro-combobox during mount).
+      // Check for any element (including custom elements like <auro-icon>
+      // that render via shadow DOM with no text content).
+      const lightDomNodes = Array.from(this.querySelectorAll('[slot="displayValue"]:not(slot)'));
+      this.hasDisplayValueContent = lightDomNodes.some((node) => (
+        node.textContent.trim().length > 0 ||
+        node.children.length > 0 ||
+        node.shadowRoot !== null
+      ));
```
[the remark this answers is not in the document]
jason-capsule42: We do not intend to support these additional element types in displayValue content so there is no need to account for them here.

components/input/src/auro-input.js
```diff
@@ -323,15 +323,44 @@ export class AuroInput extends BaseInput {
...
+    // treated as content even when they have no text or light-DOM children.
+    this.hasDisplayValueContent = nodes.some((node) => {
+      if (node.nodeType === Node.TEXT_NODE) {
+        return node.textContent.trim().length > 0;
+      }
+      if (node.nodeType === Node.ELEMENT_NODE) {
+        return node.textContent.trim().length > 0 ||
+          node.children.length > 0 ||
+          node.shadowRoot !== null;
+      }
```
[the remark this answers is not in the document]
jason-capsule42: We do not intend to support these additional element types in displayValue content so there is no need to account for them here.

components/dropdown/src/auro-dropdown.js
```diff
@@ -677,15 +719,17 @@ export class AuroDropdown extends AuroElement {
     }
 
     const eventType = event.detail.eventType || "unknown";
-    if (!this.isPopoverVisible && this.hasFocus && eventType === "keydown") {
-      this.trigger.focus();
+    if (!this.isPopoverVisible && this.hasFocus && eventType === "keydown" && !this.noHideOnThisFocusLoss) {
+      this.focusTrigger();
     }
```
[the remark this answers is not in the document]
jason-capsule42: No, you should not make this change. The reviewer's concern is theoretical but doesn't apply in practice:

The combobox already handles focus for every close path:

Escape → keyboard strategy calls [setTriggerInputFocus()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) before [hideBib()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
Selection (click/Enter) → [setClearBtnFocus()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) with the focus-chain priming
Tab → keyboard strategy's [selectAndClose()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) / [selectAndExitBackward()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
[noHideOnThisFocusLoss = true](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) semantically means "I'm managing focus/close myself" — suppressing both auto-close AND generic focus restoration is the correct interpretation of that contract.

No other consumer currently uses [noHideOnThisFocusLoss](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) without managing its own focus. The only consumer is the combobox.

You can resolve this thread by explaining that the combobox's keyboard strategy and [setClearBtnFocus](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) explicitly handle focus for all close paths — the dropdown's generic restoration was racing with them, which is why it's suppressed.

components/dropdown/src/auro-dropdown.js
```diff
@@ -69,6 +68,15 @@ export class AuroDropdown extends AuroElement {
...
     this.noHideOnThisFocusLoss = false;
 
+    /**
+     * When true, the dropdown skips its generic focus restoration on close.
+     * Set by consumers (e.g. combobox) that manage their own focus routing
+     * via setClearBtnFocus / setInputFocus / keyboard strategy.
+     * Separate from noHideOnThisFocusLoss (which controls auto-close behavior).
+     * @public
+     */
+    this.noFocusRestoreOnClose = false;
```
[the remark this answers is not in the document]
jason-capsule42: Incorrect, it is marked as @private

components/dropdown/src/auro-dropdown.js
```diff
@@ -57,7 +57,6 @@ export class AuroDropdown extends AuroElement {
   static get shadowRootOptions() {
     return {
       ...AuroElement.shadowRootOptions,
-      delegatesFocus: true,
```
jordanjones243, line 60: This is the highest-blast-radius change in the PR, and it's only validated through combobox. `auro-dropdown` is also consumed by **select, datepicker, and counter-group** — none of which have test changes here. The new `focusTrigger()` + `noFocusRestoreOnClose` compensate, but only combobox sets `noFocusRestoreOnClose`, so the other three now fall through the new `focusTrigger()` restoration path instead of the old `delegatesFocus`-backed behavior.

That's exactly the kind of change that stays green in the combobox suite and quietly regresses keyboard focus-return elsewhere. Before merge, could you confirm you've manually smoke-tested close/focus-return on select, datepicker, and counter-group (keyboard Escape + click-outside)? Not asking for new automated tests here — just a manual pass on the shared consumers.

jordanjones243 (commented):
**One follow-up (non-blocking)**
- The two original user-facing regressions still aren't directly asserted (bib opens on first keystroke after mount; overlay populates on preset value). Everything else is well covered — fine as a follow-up ticket rather than a blocker.

components/dropdown/src/auro-dropdown.js
jason-capsule42: Yes, all 4 components have been thoroughly vetted in localhost and Flight Search.

jordanjones243 (approved):

